PR #1, opened 2018-05-08
Update to Inferno

jricher:
updated configurations to use inferno paths instead of crucible_smart_app

Removed unicorn from dev image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,5 @@ FROM ruby:2.5
 COPY Gemfile* ./
 RUN bundle install
 
-# Set up unicorn (BUT the config will come from the mounted volume)
-RUN gem install unicorn
-RUN mkdir /var/unicorn && mkdir /var/unicorn/pids && mkdir /var/unicorn/log
-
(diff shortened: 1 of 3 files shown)


arscan:
This branch seemed to work on my local windows dev machine, but I'm not that familiar with this setup so I'd appreciate it if @srmoore could review and make sure there isn't anything missing.

I did notice a reference to crucible_smart_app in the nginx.conf file, though I don't know if this impacts the docker deployment.  We are going to have to do a full scrub of all documentation and deployment configuration to make sure all the directories are properly updated.  I'm ok doing that after these PRs get in so that we can get the docker part properly set up for @nbashyam though.

https://github.com/siteadmin/inferno/blob/update-to-inferno/deployment-files/nginx.conf#L49

jricher:
@arscan the docker deployment is not affected by that line, it was leftover from other configuration attempts that didn't pan out and can be cleaned up. NGINX doesn't use that directive in this setup anymore. I stopped short of doing the full scrub and just did what was needed to get the functionality going.

PR #4, opened 2018-05-31
Adding sample test data for FHIR servers

aviars:


nbashyam:
Finished merge.

jricher:
Why was this sample data put into the inferno site overlay? It should be removed from here and put into its own repository.

aviars:
I think its fine where it is for now. Rob was okay with it here.

PR #10, opened 2018-08-17
add manual confidential client

radamson:
Allows users to use register confidential clients manually or dynamically.  Added some tooltips on the view for it as well describing it.


arscan:
You also need to add very similar (or exactly the same) logic to provider ehr launch sequence.  To test that against HSPC, manually register a client (the HSPC sandbox doesn't allow you to do the EHR launch from a dynamically registered client), and choose settings on your app, and click 'Launch'.

There is massive overlap between EHR and standalone sequences.  EHR launch is basically standalone but that is initiated from a link in the EHR to this app.

arscan:
Noticed that one of the labels didn't match the checkbox, and while I was in there I fixed URI/URL inconsistencies and made the order of the 'Confential Client' be the same across modals.

PR #14, opened 2018-08-21
When executing tests, make sure to leave off trailing slash.

arscan:
Previously, when executing tasks, we added a trailing slash when hitting the URL that ran the tests.  There was a failure case where the user manually closed the modal (going to fix that in a separate PR), the user was left at a URL like `.../Conformance/`.  When clicking on the `rerun` button, it would just append `Conformance` to the previous URL, so it now hit `.../Conformance/Conformance`.

This fixes that by simply sending the user to `../Conformance` instead, so using trailing slashes doesn't break anything.

You can test by allowing the SSL timeout issue to manifest itself, then click 'off' the modal, then click `Rerun`.  This fixes that problem, and shouldn't cause any other problems.


radamson:
Fixed a merge conflict caused by another recent pull request

PR #28, opened 2018-09-04
X-Browser Bugfixes

notpace:
- [x] Fix broken regex for IE/Edge on Windows
- [x] Find and fix other IE/Edge issues on Windows
- [x] Test Safari on macOS
- [x] Test Firefox on Windows
- [x] Test Firefox on macOS
- [x] Test Chrome on macOS
- [x] Test Chrome on Windows

### DIFF
--- a/public/js/app.js
+++ b/public/js/app.js
@@ -24,33 +24,15 @@ $(function(){
    }
   });
 
-
-  // $('.sequence-main').on('click', function(e) {
-  //   if(e.target.getAttribute('role') !== 'button' && e.target.className !== 'result-details-clickable'){
-  //     $(this).parent().find('.collapse').collapse('toggle');
-  //   }
-  // });
-
   $('.sequence-expand-button').click(function (event) {
-    event.preventDefault ? event.preventDefault() : (event.returnValue = false);
-    let button = $(this)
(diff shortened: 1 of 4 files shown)


public/js/app.js
```diff
@@ -32,7 +32,7 @@ $(function(){
   // });
 
   $('.sequence-expand-button').click(function (event) {
-    event.preventDefault();
+    event.preventDefault ? event.preventDefault() : (event.returnValue = false);
```
arscan: Should we do this instead?  https://github.com/siteadmin/inferno/compare/development...a-to-button-example

I didn't merge because of conflicts.  If you agree, just grab that code and add it in here?  Anchor tags shouldn't be used for things that aren't external links.

notpace:
For future reference, here's the generic test script/things I am looking for during x-browser testing:

notpace:
Edge now works reasonably well, but IE11 has issues with bootstrap, jquery, popper, tether, open-iconic, and the custom js written for Inferno.  We'll need some polyfills and some time to test IE11 more deeply.

arscan:
Looks good, we'll hold off on IE11 and look at potentially supporting that in later versions.

PR #29, opened 2018-09-05
Detailed sequence-level explanations

arscan:
Currently just has detailed information about conformance right now.  We may consider merging this in early though to try to reduce merge conflicts later.


notpace:
Agreed

PR #32, opened 2018-09-05
TLS timeout quickly

radamson:
TLS Tester which uses NET::HTTP and times out instead of hanging on creating a TCP Socket.

Started to a unit test but I'm not sure WebMock can adequately handle testing TLS versions as SSL seems out of their purview (https://github.com/bblimke/webmock/issues/757).

We could probably do something fancier in the future using webmock and an external service (although I'm sure there are other ways as well).  I've seen some things similar to this:

```
WebMock.disable_net_connect!(allow_localhost: true)
stub_request(:any, "https://www.example.org/").to_rack(
        Rack::Handler::WEBrick.run TlsTestServer, {:SSLVersion => OpenSSL::SSL::TLS1_2_VERSION, :Port => 8000}
)
```

Default Timeout for opening the connection is 60 seconds, but we could set it lower. (See the open_timeout attribute of NET::HTTP)


notpace:
image

PR #33, opened 2018-09-06
Implement guided mode

schreiaj:
The UX for the happy path is in there

The new way to stream results back to the client is also included.


arscan:
As discussed, we are tabling this, for now at least.  Will perhaps revisit some UI elements, but probably not this implementation.  If we are going to make upgrades to page delivery we will probably go full json-api route.

PR #36, opened 2018-09-11
Modular Sinatra and Namespacing

radamson:
- [x] Update view components to handle new namespacing (e.g. "Inferno::Sequence::ExampleSequence"
- [x] Update and test rake tasks for new namespace and structure
- [x] Update and test tests for new namespace and structure


radamson:
screen shot 2018-09-11 at 6 54 27 pm
Tests pass

PR #37, opened 2018-09-12
Update base to go to index page

radamson:


arscan:
Did you check to make sure that this still works with `interno-site-overlay`?

radamson:
Tested it independently and with the site-overlay without issue

PR #40, opened 2018-09-14
Autorun conformance for newly created FHIR Servers

schreiaj:
Does support launching other tests as well as rerunning tests (so you could send someone a direct link to the test execution you're doing and it would run just as it did on your machine when you clicked the button)

autorun

### DIFF
--- a/lib/app/endpoint/home.rb
+++ b/lib/app/endpoint/home.rb
@@ -28,7 +28,7 @@ module Inferno
           url = url.chomp('/') if url.end_with?('/')
           @instance = Inferno::Models::TestingInstance.new(url: url, name: params['name'], base_url: request.base_url)
           @instance.save!
-          redirect "#{BASE_PATH}/#{@instance.id}/?autoRun=ConformanceSequence"
+          redirect "#{BASE_PATH}/#{@instance.id}/"
         end
 
         get '/:id/test_result/:test_result_id/?' do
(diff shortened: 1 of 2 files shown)


arscan:
A couple of problems:
* It will auto-scroll the page down to the conformance anchor, which loses the stuff on top.  I think this is disorienting for first visit
* Clicking the back button to get back to /inferno no longer works, because the page auto-redirects you forward.

Also, now that I see it, I'm not a huge fan of having all the tests run and show the conformance results, because then you hvae to scroll "way" down to see all the other sequences.  I think its useful to go to the page and have enough space on the page to see that that are quite a few tests (you get an overview before drilling in), instead of it getting dominated by one sequence that has already run.  (note, use an 'average' resolution to see this... obv it isn't a problem on 4k screens, but we shouldn't be optimizing for that).

Thoughts?

schreiaj:
I had a similar concern about the back button. I think I can remove the auto run state from browser history. I’ll dig into it. 

Short of having the conformance auto collapse the desire to show both those things doesn’t seem possible to do. We have to show one or the other. Or the test results. We’re gonna have to make a call on what’s priority here

I try to keep my windows small so I’m not testing on huge resolutions. I’ll do some testing with 1024x768 though.

arscan:
Yeah I don't know if there is a solution that doesn't involve tradeoffs.  All solutions I can come up with have problems (e.g. don't autoshow results breaks consistency, autoshowing results basically hides all but one sequence from a new user until they scroll, not autorunning forces user to do a click when we know they should be clicking).  Personally I'm leaning to not autorunning, making users click may be good to train them anyhow.

schreiaj:
I’d be inclined to agree. 

I’d suggest keeping the option to auto run in but disable by default. 

I’ll fix the history issue.

PR #44, opened 2018-09-21
Sequence metadata

arscan:
Refactor test metadata so that we have space for a lot more content in the interface.


radamson:
screen shot 2018-09-21 at 2 28 05 pm

arscan:
Fixed these three issues, thanks!

PR #50, opened 2018-09-25
Add optional indicator

notpace:


arscan:
I think this is fine for now.  We may be able to make these a bit more aesthetically pleasing though.

PR #56, opened 2018-09-28
Fix issue when server states OAuth endpoints but is not explicit about smart support.

arscan:
This is a very frequent issue -- the server provides OAuth endpoints but isn't explicit about SMART support.  We updated the client so that we can be relaxed about this issue (note the 3.0.6 client bump).

Perhaps we should add a warning about this?  Notably, the SMART r2 reference server has this issue.


arscan:
Added the warning.

PR #63, opened 2018-10-01
add sticky client info panel and some subtle inferno branding

schreiaj:
Client Info bar sticks to top of screen on scroll. 

 
spinny-nav


arscan:
There are couple more issues:

- [ ] for long server urls, the server url on top now wraps.  We have code in there to ellipsisize long server urls, but the changes here break that
- [ ] After you run the test, the icon doesn't appear until you scroll a bit.  It should just show up by default.

Really cool though.  Just need to tie up on the loose ends.

arscan:
While reviewing I found a few issues, so I just took care of them:
* the transitions were removed when you readded the ellipsis
* I added a link back to the base directory for the logo
* I disabled the transition when you start in the middle of the page.

Last 2 things we need to do:

- [ ] Check the in IE edge
- [ ] Try out inferno-site-overlay and update any css / html if necessary

PR #70, opened 2018-10-02
Docker config bug fix

arscan:
The docker setup for this repository had an issue where it would not launch properly.


radamson:
Tested on MacOS and Windows with docker

PR #74, opened 2018-10-03
Conformance warning

arscan:
The second warning is incorrect... the data is actually there.  Fixing this makes that warning go away if it is correct.

screen shot 2018-10-02 at 11 23 35 pm


lib/app/sequences/01_conformance_sequence.rb
```diff
@@ -173,7 +173,7 @@ class ConformanceSequence < SequenceBase
         assert (token_url =~ /\A#{URI::regexp(['http', 'https'])}\z/) == 0, "Invalid token url: '#{token_url}'"
 
         warning {
-          service = @conformance.try(:rest).try(:first).try(:security).try(:service).try(:coding).try(:code)
+          service = @conformance.try(:rest).try(:first).try(:security).try(:service).try(:first).try(:coding).try(:first).try(:code)
```
jmandel: These "first"s aren't precisely correct; if you come across a server that supports more than one security protocol, suddenly there will be an order dependency whereas the fhir specification would consider the order of multiple entries irrelevant.

Perhaps the philosophy is just to wait until it matters for a real user and fix it then (which is fine -- just want to make sure it is an intentional choice).

radamson: Updated to check all security services, so it will work with conformance statements with multiple entries:

screen shot 2018-10-03 at 11 10 10 am

Updated error messages to show what was found (for example with an accidental misspelling)

screen shot 2018-10-03 at 11 12 23 am

PR #75, opened 2018-10-03
Add observation codes valueset

jmandel:
I know these files aren't doing anything yet, but the other valuesets are represented already, Observation.code and Condition.code are important bits of structured data -- plus, this is getting things set up for being able to validate codes better.

source: http://hl7.org/fhir/DSTU2/valueset-observation-codes.canonical.json


radamson:
Closing this because we added the Observation codes in a later PR that updated the validator.  Thanks for the PR!

PR #80, opened 2018-10-05
Argonaut valueset checks

jmandel:
This is a proof of concept to add valueset membership checks as part of argonaut profile validation. It's intended to be used with a pack of bloom filters for key vocabularies and valuesets -- so we can distribute a set of files for validation without having to actually distribute vocabulary content (doing so directly can be large/unwieldy, and can come with licensing concerns).

Some notes:

* I haven't really written (or read much!) ruby, so I'm sure there's a lot of non-idiomatic stuff here

* Partly I needed to add functionality to the underlying `fhir_dstu2_models` library, but I didn't know a good way to do this in a PR across repos so I just added the one file I wanted to tweak here in this repo, and show how to mount it in with a docker volume map to overlay it during dev/testing. I'd love to know the "right" way to do this in dev, and in a PR, but I was working from an airplane without network access, so I just threw this together. In any case, https://github.com/siteadmin/inferno/commit/c7d8f8285de207226735740041ed1f50b0d54adb shows the precise _changes_ to the file from `fhir_dstu2_models`

* Rather than looking over the full diff (two commits), it'll be easier to look at https://github.com/siteadmin/inferno/commit/c7d8f8285de207226735740041ed1f50b0d54adb directly, since it highlights the changes to the underlying validation library.

* It looks like only validation errors are returned from the profile validation step; it may make sense to return warnings too (though there's also an open question of whether to highlight vocab mismatches for, say, an extensible binding as *errors*, *warnings*, or something else. My preference is to make sure they're at least *somehow* visible.

Anyway, I'd be very pleased to see this functionality incorporated into Inferno -- whether you'd want to use this PR as inspiration for a proper implementation, or if you want to guide me in the right direction.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,5 +20,6 @@ gem 'json-jwt'
 gem 'colorize'
 gem 'kramdown'
 gem 'selenium-webdriver'
+gem 'bloomer'
 
 gem 'simplecov', require: false, group: :test
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,9 @@
 version: '3'
 services:
   # the main ruby server
+  smart_server:
+    image: syncforscience/reference-api
+
   ruby_server:
     # build the image from ./Dockerfile (this installs the dependencies into the image)
     build:
@@ -8,6 +11,7 @@ services:
     # map the project directory into the running container
     volumes:
       - .:/var/www/inferno
+      - ./structure_definition.rb:/usr/local/bundle/gems/fhir_dstu2_models-1.0.4/lib/fhir_dstu2_models/fhir_ext/structure_definition.rb
     # expose the internal application to direct HTTP requests
     #ports:
     #  - "4567:8080"
--- a/download-argonaut-valuesets.sh
+++ b/download-argonaut-valuesets.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+wget https://github.com/jmandel/umls-bloomer/archive/bloom-filters.zip
+unzip bloom-filters.zip -d data
+rm bloom-filters.zip
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -6,8 +6,33 @@ require_relative 'utils/exceptions'
 require_relative 'utils/validation'
 require_relative 'utils/web_driver'
 
+require 'bloomer'
+require 'bloomer/msgpackable'
+require 'json'
+
 module Inferno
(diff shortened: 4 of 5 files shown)


jmandel:
@steve-posnack FYI, since we discussed terminology validation at the HL7 meeting this week

radamson:
@jmandel we'd like to pull in your bloom filter approach so that users can create the validators with their own UMLS login credentials.  We've been pulling your contributions and some other functionality together on [an updated branch](https://github.com/siteadmin/inferno/compare/termin), but would like to include [your code for generating the filters](https://github.com/jmandel/umls-bloomer/blob/master/04-generate-bloom-filters.rb) in too.  Are you ok with us using it in Inferno or would you be willing to submit a PR with that file to recognize your contribution (which is greatly appreciated)?

jmandel:
Just noticed that I missed this comment! I would be happy to submit a PR if you let me know which branch or repository to do it against.

arscan:
Hi @jmandel -- we followed your proof-of-concept and integrated it into here (and the underlying models library) already.  Thanks for the direction!  A couple of notes:

* We provide a couple of options for testing set inclusion, but the default configuration is to use the bloom filter as you had done.  It's pretty remarkable how memory .efficient it is.
* We return vocab mismatches as warnings for extensible bindings, instead of errors
* In the case where there is a vocab mismatch on the ValueSet for extensible bindings, we attempt to see if the code is in the stated CodeSystem and return a warning if it isn't (assuming we are aware of the CodeSystem).
* We don't distribute the validation packs, but [provide scripts](https://github.com/siteadmin/inferno/wiki/Installing-Terminology-Validators) that pull down codes from UMLS and process them.  We followed your approach on this as well.

I'm going to close this because all the functionality you provided here is now integrated.  Thanks again!

jmandel:
Phew -- I'm just catching up on this and delighted to see how much of my initial sketch you've incorporated!  If you don't mind answering here in this closed issue @arscan, do the Inferno editions at https://inferno.healthit.gov/ include this behavior, and do they use the (default) bloom filter approach for it?

arscan:
Yes, we are using bloom filters on the inferno.healthit.gov deployment (both community and program) to check valueset bindings.  We are currently only checking terminology on the Argo DSTU2-based tests -- our US Core R4 tests are a work in progress and we haven't gotten to terminology yet. Should be straightforward though, and I expect we'll have time to do that soon.

Thanks for providing direction and kickstarting our implementation @jmandel!

PR #82, opened 2018-10-09
Safeguard Conformance Security Service Checks

radamson:
### Protects against empty/missing security elements in rest elements.

This prevents failure caused by rest elements which do not include any security elements as defined in the [DSTU2 Conformance Documentation](https://www.hl7.org/fhir/DSTU2/conformance.html) : 

screen shot 2018-10-09 at 10 04 44 am

For example, if the JSON array of rest elements in the conformance statement contains an empty rest element: 

screen shot 2018-10-09 at 9 17 31 am

Then it will not issue a failure or warning about the supported security service as long as one of the rest elements contains the correct ["SMART-on-FHIR' security service code](http://docs.smarthealthit.org/authorization/conformance-statement/)


arscan:
I'm a little concerned about mixing the `try` and the `&.` syntax but we can have it here and look to clean this up globally later.

PR #83, opened 2018-10-11
Update client state view

radamson:
Minor Updates to the client state view:

- Remove "Conformance Statement Resource Support" before supported resources have been identified
- Add client secret when the instance is used as a confidential client
- Removed some unnecessary commas

### DIFF
--- a/lib/app/views/state_status.erb
+++ b/lib/app/views/state_status.erb
@@ -7,17 +7,16 @@
       </div>
     </div>
 
-    <h4>Conformance Statement Resource Support</h4>
 
-    <% if instance.supported_resources.count %>
-      <ul>
+    <% if instance.supported_resources.count > 0%>
+    <h4>Conformance Statement Resource Support</h4>
+    <ul>
       <% instance.supported_resources.each do |resource| %>
         <% if resource.supported %>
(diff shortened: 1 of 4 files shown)


lib/app/views/state_status.erb
```diff
@@ -7,17 +7,16 @@
...
+    <h4>Conformance Statement Resource Support</h4>
+    <ul>
       <% instance.supported_resources.each do |resource| %>
         <% if resource.supported %>
           <li><%= resource.resource_type %> [
               <% if resource.read_supported %> read, <% end %>
               <% if resource.vread_supported %> vread, <% end %>
-              <% if resource.history_supported %> history, <% end %>
-              <% if resource.scope_supported %> authorized <% end %>
+              <% if resource.history_supported %> history<% end %><% if resource.scope_supported %>, authorized <% end %>
```
arscan: Hmm, this will result in weird behavior if vread and scope are supported but not history.  I think we should add a method to the resource model that is something like 'supported_interactions', which returns an array and we can just `join(',')` out here.

radamson: I like that idea, hopefully you like the implementation

PR #84, opened 2018-10-17
Conformance Supports Check

schreiaj:
Adds a warning to users if the Conformance doesn't support the required resource for a sequence. 

Doesn't currently check for operations on those resources. 

screen shot 2018-10-17 at 3 51 17 pm

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -18,7 +18,7 @@ disable_verify_peer: false
 bind: '0.0.0.0'
 
 # Skip TLS tests
-disable_tls_tests: false
+disable_tls_tests: true
 
 # Default scopes
 default_scopes: launch launch/patient offline_access openid profile user/*.* patient/*.*
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -31,6 +31,7 @@ module Inferno
       @@descriptions = {}
       @@details = {}
       @@requires = {}
+      @@conformance_supports = {}
       @@defines = {}
       @@test_metadata = {}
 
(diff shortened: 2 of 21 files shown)


radamson:
Nice, I like it.

Just testing it out and noticed a couple things:

I wasn't able to get the warning to appear on any of the `Authentication and Authorization` sequences like `Dynamic Registration` in your example.  Maybe just a missing commit?

Also interestingly, I see the warnings as green in Chrome:

screen shot 2018-10-17 at 5 59 00 pm

Might be nitpicking, but maybe say "The Conformance Statement does not indicate support for the Resources required for this test" instead of "Your Conformance says you..."

arscan:
I'm not sure if using that space is the right place to display this warning, as it implies (to me at least) that something has been run.  Also, it would be replaced with a result the user does run something.

schreiaj:
@arscan That was actually intentional. Once they've run a test the warning of "hey you don't say you support this are you sure" is kind of silly. Either the tests passed and their conformance is wrong or they failed and it's because they don't support it. 

@radamson It's based off of a new field in the sequence metadata specified like `conformance_supports :AllergyIntolerance` I pulled these all out of the tests with `skip_if_not_supported` which really just means the argonauts tests. I'll look into why it's showing up as green, possibly I forgot to change some CSS. 

The reason the example is the Dynamic Registration is because it was easy to test logic there since it was almost always in view on page load. I'd manually added `conformance_supports :ResourceThatDoesntExist` to trigger the logic.

arscan:
I think this should be set up so that the visual indicator is only visible after you run the conformance statement.  Since we don't know one way or another yet, I think we should assume it is supported until they tell us it isn't.

Also, I think moving the indicator over to near the run button somehow, plus maybe muting the row a bit, might be a good approach.  Could you try that out?

schreiaj:
Crap, I seem to have screwed up a rebase somewhere along the line. I'll fix this in the AM.

schreiaj:
Updated to not show warning after run. 

Also removed skip if not supported behavior since it's superseded by warning them that stuff may break.

PR #94, opened 2018-11-06
Add reset task, also add cron dependency for use on the demo server. 

schreiaj:


arscan:
I wonder if this could cause any problems depending on the order these deletions are run.  

Could you do `DataMapper.auto_migrate!`

Looks like that drops and recreates all table definitions, which seems like it might be a more foolproof method: https://datamapper.org/getting-started.html

PR #99, opened 2018-11-09
MedicationReference Checking

radamson:
- https://github.com/siteadmin/inferno/commit/d003063fe1b0e42d40a2d4bdf227e5bee272ba03 contains style and documentation updates
- https://github.com/siteadmin/inferno/commit/4274b42cdd6f52efe0aa3a251bc10f9357ad6bf8 makes existing MedicationStatement tests go over all MedicationStatements, not just the first
- The rest focuses on reading MedicationReferences provided and evaluating them against the relevant Argonaut Profile

Looking for feedback on: 

- `check_resource_against_profile` vs the existing method `test_resources_against_profile`.  The existing method retrieves references from the database and reads them, which might not be possible with contained resources [1].  I'm sure my implementation could be improved (and the commonality better considered).

- `validate_reference_read_reply` vs `validate_read_reply` for similar reasons (and the fact that the `id` property of the reference doesn't align with the id of the resource referenced

[1] "In some circumstances, the content referred to in the resource reference does not have an independent existence apart from the resource that contains it - it cannot be identified independently, and nor can it have its own independent transaction scope." - [DSTU2 Contained References](https://www.hl7.org/fhir/DSTU2/references.html#contained)


arscan:
I added a `pass` option that allows the user to return a message on a pass.   Also a few cleanups.  Are you ok with my changes @radamson

PR #108, opened 2018-12-06
WIP: Update Rake and CLI

radamson:
Updates to rake tasks and CLI

- [x] Update `tests_to_csv` export
  - [x] Add test for `tests_to_csv`
- [x] Update `generate_config` task
- [x] Update `execute` task
- [x] Update `execute_batch`

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -83,7 +83,7 @@ GEM
       multipart-post (>= 1.2, < 3)
     fastercsv (1.5.5)
     ffi (1.9.25)
-    fhir_client (3.0.6)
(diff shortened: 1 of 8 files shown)


arscan:
I updated the rake tasks to support the refactored code.  The one requirement is that we now need to pass in a module for this to work because that is where we discover which version of FHIR we are using.  For the batch stuff, I'm not sure how much sense this makes.  I considered having dstu2 and stu3 placeholder modules that contain all sequences that support each version.  But I don't think we need to necessary have that settled quite yet.

PR #120, opened 2019-01-28
UI improvements

arscan:


arscan:
FYI @radamson @czh-orz -- I'm putting in a fairly large UI update in development that has incorporated your changes up to this point.

PR #125, opened 2019-02-06
Report fixes

czh-orz:
Changed report UI. 
Show optional results in test cases. 
Find request count and last sequence time via database.
Fix redirect URI in client state


arscan:
Looks good!  Two things I think could be improved:

1) Still showing skipped (see below screenshot).  Just condense all statuses to the concept of passed vs not passed... skipped counts as not passed.
2) Since we have space, lets be more clear what we mean, `4/5 Required Tests Pass - 1/2 Optional Tests Pass`

showing_skipped

arscan:
Looks good, thanks!

PR #128, opened 2019-02-12
Rerun skipped

czh-orz:
Added button to rerun skipped tests for each group. 
I'm not sure if I should fix the launch/patient scope issue (during second launch attempt) in this PR. Let me know if I should.

### DIFF
--- a/lib/app/module.rb
+++ b/lib/app/module.rb
@@ -49,8 +49,9 @@ module Inferno
       attr_accessor :id
       attr_accessor :test_cases
       attr_accessor :run_all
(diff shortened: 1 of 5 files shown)


arscan:
Let's keep the `launch/patient` issue separate, as it is independent of this.  Show me that issue when I'm in the office next.

arscan:
:100 this is awesome.  Much cleaner than I expected it to be!  We probably should add language about how you don't have to have all required Resources within a single launch.  But I think it is probably fairly self evident as-is.  Great job!

PR #132, opened 2019-02-13
Test logs

czh-orz:
Add log tab in each test case.
Currently, the log only shows request responses.


arscan:
Looks great!

Made some minor changes: eventually this tab will include much more detailed logs, but for now it is just requests/responses, so I changed the tab name to `HTTP Request Log`.  I also change where the details link shows up to just be consistent with the other tab.

PR #133, opened 2019-02-14
Adding details on sequences and sequence groups.

radamson:
Adds overview text to sequences and details to each group.  The group details are now also displayed in the developer view.

screen shot 2019-02-14 at 11 00 05 am


arscan:
Fixed page breaking if no overview provided.  Also I removed the overview content for now in the developer view module, because it will require some styling to look good, and that is a lower priority at the moment.

arscan:
Note, this is just a first pass, we'll be making more updates to documentation.

PR #136, opened 2019-02-18
show uris during registration sequences

czh-orz:
I'm leaving it specific to URIs right now, but it should be easy to expand this to be generic in the future if needed.


arscan:
Looks great, and works in both views.

I was wondering if 'show_uri' is a little too specific to those variables, and if we should instead have it be a bit more generic, but those variables are somewhat unique (since they are provided by Inferno), so I think this is a good place to start and if we have other things that should show up like in the future then we can make it more generic.

PR #137, opened 2019-02-19
Scope validation

czh-orz:
Validate scopes specifically for ONC context.


arscan:
Looks good.  I added functionality to our modules so you can specify default values on variables, which is nice because we can set different default scopes on each of the launches.  The one downside to this right now is that you used to be able to check what scopes were used last time when starting a new test, but now it just overwrites them.  We can see how annoying this is in the future and perhaps update it. But I think it is an improvement at least for now.

I also tweaked the tests slightly.

PR #138, opened 2019-02-19
Fix transferring typo.

arscan:


arscan:
This was fixed in another PR.

Issue #139, opened 2019-02-19
Client secret doesn't save during the Manual Registration sequence for confidential clients 


abjonnes:
When running the Manual Registration sequence for a confidential client, only the client ID gets saved to the client state.

arscan:
Thanks for finding this issue and submitting a potential fix.  This should be all set in `master` and on https://inferno.healthit.gov.  I'll go ahead and close this but let us know if you are still seeing the problem on your side.

PR #140, opened 2019-02-19
Save client secret to instance state during Manual Registration sequence

abjonnes:
Quick fix for #139. A more robust solution might be to recurse through any child `div.form-group`s and collect all the `data-prerequsite`s that way.


arscan:
Thanks @abjonnes !  @czh-orz could you take a look at this?

arscan:
Thanks again @abjonnes!  We ended up going with a slightly different fix in #146 that is a bit more consistent with our current approach.  This should be all set on https://inferno.healthit.gov and in `master`.

PR #141, opened 2019-02-20
overflow the response values

czh-orz:


arscan:
Are there any downsides to doing this?

czh-orz:
I don't think so. It was overflow-x: scroll before. The difference between auto and scroll is that auto will only show the scrollbar if content is being clipped. And I also don't specify the x axis since that caused issues with the y scrollbar for me.

PR #143, opened 2019-02-21
separate client confidential and client secret prerequisites

czh-orz:
save client secret during registration by moving it out of the confidential_client prerequisite field.

### DIFF
--- a/lib/app/modules/smart/dynamic_registration_sequence.rb
+++ b/lib/app/modules/smart/dynamic_registration_sequence.rb
@@ -31,8 +31,8 @@ module Inferno
 
       show_uris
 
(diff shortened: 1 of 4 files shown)


lib/app/modules/smart/dynamic_registration_sequence.rb
```diff
@@ -31,8 +31,8 @@ class DynamicRegistrationSequence < SequenceBase
 
       show_uris
 
-      requires :oauth_register_endpoint, :client_name, :initiate_login_uri, :redirect_uris, :scopes, :confidential_client,:initiate_login_uri, :redirect_uris, :dynamic_registration_token
```
arscan: This is incorrect -- for Dynamic Registration we need to indicate whether we are registering a confidential or public client, but we do not "input" the client secret.  Instead, we receive the client secret from the dynamic registration server. Could you revert this and let me know if your solution still works?

czh-orz: Reverted change and the solution still works.

arscan:
Looks good 👍

PR #148, opened 2019-02-25
Presets with s4s demo

abjonnes:
Hi again,

I've put together a small example of some additional functionality that we need in a test suite for use at S4S: the ability to store configurations for vendors, and to complete the authorization process in an automated fashion. I've included a configuration file for the [S4S reference implementations](https://github.com/sync-for-science/reference-stack-docker). I've also made a few rough changes to get the headless Chrome instance running in the Docker container.

### Vendor configurations
Vendor configurations are the same format as the existing batch script configuration with the addition of an optional `preset_name` in the root object. I've stored these "presets" in a `presets/` directory. The directory is scanned for the configuration files, and a dropdown is added to the home page allowing the user to select one.

image

When a preset is selected, Inferno populates the interface with the vendor's URI. Upon submission, the new testing instance will be populated with the data from the configuration.

image

The rest of the Inferno workflow is mostly unchanged, except for the registration and standalone launch sequences.

### Automated authentication
Since the testing instance now includes the `standalone_launch_script` from the configuration, Inferno automatically uses the script during the standalone launch sequence, requiring no input from the user.

---

If your team thinks this type of functionality is useful, S4S would be happy to collaborate further to refine and extend it. (This is my last week at S4S, but @jmandel and my eventual replacement will continue to work on adapting Inferno to meet our test suite needs!)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,3 +6,12 @@ RUN bundle install
 
 # Expose the port
 EXPOSE 4567
+
+RUN apt-get update
+RUN apt-get install -y libnss3 fonts-liberation libappindicator3-1 libasound2 libatk-bridge2.0-0 libgtk-3-0 libnspr4 libx11-xcb1 libxss1 libxtst6 lsb-release xdg-utils
+
+# Install Chrome
+RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+RUN dpkg -i google-chrome-stable_current_amd64.deb; apt-get -fy install
+
+ADD ./chromedriver /usr/local/bin
--- a/lib/app/endpoint/home.rb
+++ b/lib/app/endpoint/home.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../utils/instance_config'
+
 module Inferno
   class App
     class Endpoint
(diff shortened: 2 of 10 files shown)


arscan:
We definitely want this functionality, thanks!  It may take a few days before we have a chance to review this though.

Question: did the chromedriver binary come from http://chromedriver.chromium.org/downloads ?

Looks like version of chromedriver depends on the version of chrome you are running, so we may need to be careful about getting the right versions (most recent chrome vs. a static copy of chromedriver). And if we already have 'downloading' in the workflow, perhaps we should download & unzip this file too, just so we don't have to redistribute it.  What do you think?

Sorry to hear you are moving on, thanks for all your contributions to this project!

abjonnes:
I fully agree with your suggestions about chromedriver! I hadn't paid much attention to that part, just threw something together quickly to get it working. Looking forward to your further feedback. 😃

arscan:
Does this include fixes for the Cerner & WebDriver issue you were observing @abjonnes?  Or is that still an open issue?

abjonnes:
It's still an open issue unfortunately. I didn't hear back from Jenni after my last email to you and her about it on January 29th. Did you get a chance to look at it yourself?

arscan:
Ok, thanks.  I have not unfortunately (sorry for that!).  I'll try to get to it as soon as I can.

radamson:
I'm going to close this PR as the code base has changed significantly since this PR and support for presets has been added in https://github.com/onc-healthit/inferno/pull/205.  There is still more that should be included relevant to this PR, including allowing presets to use WebDriver.

I opened a ticket (FI-382) for including WebDriver support and we can reopen this PR if there is any new activity updating this at that time.

PR #154, opened 2019-03-01
Extend search validation

czh-orz:
I'm adding a validate_resource_item function in each sequence that does a validate_search_reply. This makes it so each sequence has a way to validate search results based on its own structure. 

I still need to implement the date check, but I'm not sure how to check if a date time falls within a period. It should be easy to add later though. Not sure if this is time-sensitive so just creating a pull request now.

Things I'm not sure about: 
It looks like clinicalStatus in the condition resource should be a code, but in HSPC it's just a string.
Argonaut says category is a valid search parameter for careplan, but FHIR DSTU2 says it isn't.

### DIFF
--- a/lib/app/modules/argonaut/argonaut_allergy_intolerance_sequence.rb
+++ b/lib/app/modules/argonaut/argonaut_allergy_intolerance_sequence.rb
@@ -13,6 +13,13 @@ module Inferno
       requires :token, :patient_id
       conformance_supports :AllergyIntolerance
 
+      def validate_resource_item (resource, property, value)
+        case property
+        when "patient"
+          assert (resource.patient && resource.patient.reference.include?(value)), "Patient on resource does not match patient requested"
+        end
+      end
+
       details %(
         # Background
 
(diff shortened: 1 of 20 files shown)


arscan:
"It looks like clinicalStatus in the condition resource should be a code, but in HSPC it's just a string."

A [code is a primitive](http://www.hl7.org/fhir/DSTU2/datatypes.html#1.19.0.1.2) that is a simple string with a few other minor constraints (e.g. no whitespace).  I think you are confusing this with the [Conditon.code](http://www.hl7.org/fhir/DSTU2/condition-definitions.html#Condition.code) element, which confusingly has the type [CodeableConcept](http://www.hl7.org/fhir/DSTU2/datatypes.html#CodeableConcept), instead of being of type `code`.

So you are basically confusing a type with the name of an attribute that has a different type.  Don't worry, I think this is confusing too.

arscan:
I think we'll need to clean up how we access codes within CodeableConcepts, instead of grabbing `first` we should look for `any?`.  Also we probably should move over to using the [Safe Navigation Operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) now that we are Ruby 2.3+.  See my last commit to see what I'm talking about.

I think we are good enough on here for now.   Good job!

PR #156, opened 2019-03-04
Report test inputs

czh-orz:
Displays inputs in both the report and as a tab in the test case.


arscan:
screen shot 2019-03-04 at 1 56 07 pm

Looks like the fields are duplicated here.  Also, variables with colons in them seem to mess up the formatting.

PR #164, opened 2019-03-08
Add inferno instance uri to report.

arscan:
It would be useful to see in the report which operational system performed the tests, and be able to get back to the live instance.  This adds the URI to the report.

It also fixes a bug where on the report section, the report stated 'Step 4 of 4' incorrectly.

screen shot 2019-03-08 at 11 38 31 am


arscan:
Good idea.
screen shot 2019-03-08 at 1 42 33 pm

PR #165, opened 2019-03-08
open failed sequences correctly

czh-orz:
To reproduce, run the standalone launch sequence in the ONC module. The failed tests should be open and not just the first sequence.


arscan:
Confirmed that it works now, good catch.

PR #172, opened 2019-03-29
Show state changes

czh-orz:
Display state changes after running sequences.


arscan:
Looks great!  The idea of 'state changes' was a little confusing because the tab didn't show up sometimes when I was expecting because nothing changed (maybe because I was rerunning the test), so I just switched it to be 'output' for now and always printed what was `defined`.  We can see how things work in practice though.

PR #193, opened 2019-05-17
R4 test generator

arscan:


generators/argonaut-r4/generator.rb
```diff
@@ -0,0 +1,504 @@
...
+
+  is_first_search = search_test[:index] == '02' # if first search - fix this check later
+  if is_first_search then
+    search_test[:test_code] = %(
+        #{get_search_params(sequence[:resource], sequence[:profile], search_param[:names])}
+        reply = get_resource_by_params(versioned_resource_class('#{sequence[:resource]}'), search_params)
+        assert_response_ok(reply)
+        assert_bundle_response(reply)
+
+        resource_count = reply.try(:resource).try(:entry).try(:length) || 0
```
radamson: Do we want to use `#try` or the `&.` here?

arscan: We should get rid of the try's, but let's split this update out into the next PR.

PR #195, opened 2019-05-24
display fhir version

nicole-kennedy:
Displayed FHIR version in the heading:
fhir version

And on the index page: 
start testing

But not for SMART: 
smart


arscan:
Could you also put this on the index page as well?

Screen Shot 2019-05-24 at 10 55 21 AM

PR #196, opened 2019-05-24
Made fhir version optional

nicole-kennedy:
The fhir_version field in the yml files for configuring modules is now optional. If this field is filled, then the FHIR version is displayed on the index page, header, and in the state popup box. For example, the SMART module does not have a fhir_version. Also, the FHIR version was removed from the module titles. 

Index page: 
start

Header: 
header on
header off

State popup box: 
state on 

state off

### DIFF
--- a/lib/app/views/default.erb
+++ b/lib/app/views/default.erb
@@ -1,7 +1,7 @@
 <div class="container" role='main'>
   <div id="header-module-name">
   <%=instance.module.title%>
-  <% if instance.module.title != "SMART Application Launch Framework Implementation Guide v1.0.0" %>
+  <% if instance.module.fhir_version != nil %>
     <span class='badge badge-dark'>FHIR <%=instance.module.fhir_version.upcase%></span>
   <% end %>
   </div>
--- a/lib/app/views/default.erb
+++ b/lib/app/views/default.erb
@@ -1,7 +1,7 @@
 <div class="container" role='main'>
   <div id="header-module-name">
   <%=instance.module.title%>
-  <% if instance.module.fhir_version != nil %>
+  <% unless instance.module.fhir_version.nil? %>
     <span class='badge badge-dark'>FHIR <%=instance.module.fhir_version.upcase%></span>
   <% end %>
   </div>
--- a/lib/app/views/guided.erb
+++ b/lib/app/views/guided.erb
@@ -1,7 +1,7 @@
 <div class="container" role='main'>
   <div id="header-module-name">
   <%=instance.module.title%>
-  <% if instance.module.title != "SMART Application Launch Framework Implementation Guide v1.0.0" %>
+  <% if instance.module.fhir_version != nil %>
     <span class='badge badge-dark'>FHIR <%=instance.module.fhir_version.upcase%></span>
   <% end %>
   </div>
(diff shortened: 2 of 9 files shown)


lib/app/views/default.erb
```diff
@@ -1,7 +1,7 @@
 <div class="container" role='main'>
   <div id="header-module-name">
   <%=instance.module.title%>
-  <% if instance.module.title != "SMART Application Launch Framework Implementation Guide v1.0.0" %>
+  <% if instance.module.fhir_version != nil %>
```
radamson: `#nil?` would be more Ruby idomatic

i.e. `unless instance.module.fhir_version.nil?`

[#nil?](https://ruby-doc.org/core-2.6.3/Object.html#method-i-nil-3F)
[#unless](https://ruby-doc.org/docs/keywords/1.9/Object.html#method-i-unless)

lib/app/views/guided.erb
```diff
@@ -1,7 +1,7 @@
 <div class="container" role='main'>
   <div id="header-module-name">
   <%=instance.module.title%>
-  <% if instance.module.title != "SMART Application Launch Framework Implementation Guide v1.0.0" %>
+  <% if instance.module.fhir_version != nil %>
```
radamson: See above about `#nil?`

lib/app/views/module_options.erb
```diff
@@ -7,7 +7,7 @@
       <input class="form-check-input" type="radio" name="module" value="<%=mod.name %>" id="<%=mod.name %>" <% if index == 0 %>checked<% end %>>
       <label class="form-check-label" for="<%=mod.name %>">
       <%=mod.title %>
-      <% if mod.title != "SMART Application Launch Framework Implementation Guide v1.0.0" %>
+      <% if mod.fhir_version != nil %>
```
radamson: See above about `#nil?`

lib/app/views/state_status.erb
```diff
@@ -7,12 +7,14 @@
       </div>
     </div>
 
-    <div class="form-group row">
-      <label class="col-sm-2 col-form-label">FHIR Version</label>
-      <div class="col-sm-10">
-        <input type="text" class="form-control" value="<%=instance.fhir_version.upcase %>" disabled>
+    <% if instance.fhir_version != nil %>
```
radamson: See above about `#nil?`

arscan:
I tried running a sequence in the SMART App Launch Guide v1.0.0 and I get an error:

Screen Shot 2019-05-27 at 10 16 06 PM

That purple "!" status means that an uncaught exception occurred during the execution of the sequence.  From our perspective, that means something is wrong on our side (vs a red "Error", which is a problem on the system under test's side).

The error message isn't very helpful there.  To see a better message, replicate the error and take a look out the console output.  There should be a "Fatal Error" line with that same message, followed by a stack trace that you can use to figure out what is going wrong.

I assume somewhere in our code we assume that the version is a string, when it could be nil, as is the case for this test module.

nicole-kennedy:
Tested Proposed ONC Health IT Certification Tests with DSTU2 and R4, looks good except for this error from R4 (also put on slack)
error

radamson:
This looks like an unrelated issue that is addressed in https://github.com/siteadmin/inferno/pull/193.

PR #197, opened 2019-05-28
WIP: Fix CareTeam test in DSTU2

arscan:
CareTeam incorrectly skips on a correctly behaving server.  This is because how we handle references and guess_profile for resources that could match multiple profiles.  I added a test that shows the issue in this PR to start.

Screen Shot 2019-05-28 at 2 24 10 PM


arscan:
Let's move the profile issue elsewhere.

arscan:
Rebased in #203

PR #202, opened 2019-05-31
Autocorrect linting errors

Jammjammjamm:
This branch fixes the autocorrectable rubocop errors and configures Travis to run rubocop as part of CI.


config.ru
```diff
@@ -1,4 +1,6 @@
-#\ -s Thin -p 4567 -q
+# frozen_string_literal: true
+
+# \ -s Thin -p 4567 -q
```
arscan: This is special `rackup` syntax that cannot have that space.  It must be `#\ -s Thin -p 4567 -q`.  I suggest we revert this change and always Exclude this file for AllCops in `rubocop.yml`.  But there could be a better way.

To demonstrate the error, when you do `bundle exec rackup` it will serve on port `9292` instead of `4567`.  I'd prefer to just keep the same port as we had before (a Sinatra default vs a Rack default I think).

PR #204, opened 2019-05-31
Fix linting errors

Jammjammjamm:
This branch contains fixes for linting errors which are trivial to fix but can't be automatically corrected. Fixing the remaining errors will require some deeper thinking.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -96,15 +96,6 @@ Metrics/ParameterLists:
 Metrics/PerceivedComplexity:
   Max: 32
 
-# Offense count: 3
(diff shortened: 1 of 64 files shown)


arscan:
I manually tested the application to make sure that this didn't cause any visible problems, and obviously it passes all of our automated tests, though I think on this upcoming release in particular we'll want to add more tests for our sequences to verify all behavior is correct.  I'm going to go ahead and merge to get all of these lint fixes in at once.

FYI @czh-orz -- this will cause your generated code to fail if you rerun the generator, as we are only fixing the generated code.  Please refer to the fixes here and have them reflected in the generator, and ask @Jammjammjamm if needed.  Just regenerating the code and doing a `git diff` highlights what is changed really well.

PR #205, opened 2019-06-03
"Try Me" FHIR Server Option

nicole-kennedy:
Added a "Try Me" option for testing an already existing and setup FHIR server. The presets are configured in the config file and include the name of the preset server, link, module it should be run with, and the fields that should be automatically filled (ex. client id, client secret). There can also be an optional link to instructions, which are kept in the Wiki. 

Screen Shot 2019-06-06 at 1 18 36 PM

Screen Shot 2019-06-06 at 1 18 48 PM

Screen Shot 2019-06-06 at 1 18 57 PM

Similar and credit to #148.


lib/app/endpoint/home.rb
```diff
@@ -11,7 +11,7 @@ class Home < Endpoint
 
         # Return the index page of the application
         get '/?' do
-          erb :index, {}, modules: settings.modules.map { |m| Inferno::Module.get(m) }.reject(&:nil?)
+          erb :index, {}, modules: settings.modules.map { |m| Inferno::Module.get(m) }.reject(&:nil?), presets: settings.presets
```
Jammjammjamm: You can use `.compact` rather than `.reject(&:nil)`

nicole-kennedy: Sounds good, the `.reject(&:nil)` was from the merge conflict so I thought that was something @Jammjammjamm changed, but I will change it to `.compact`.

Jammjammjamm: Yeah, rubocop knows to change instances of `.select { !something }` to `.reject { something }`, but it doesn't know to change `.reject(&:nil?)` to `.compact`.

PR #206, opened 2019-06-04
Check for valid JSON

nicole-kennedy:
Added checking for valid JSON to smart sequences token refresh, EHR launch, and standalone launch sequence. 

Screen Shot 2019-06-04 at 1 32 37 PM
Screen Shot 2019-06-04 at 1 32 44 PM


arscan:
Looks great!

PR #207, opened 2019-06-04
Minor refactors

Jammjammjamm:
Some more minor refactoring.

### DIFF
--- a/lib/app/models/sequence_result.rb
+++ b/lib/app/models/sequence_result.rb
@@ -58,6 +58,17 @@ module Inferno
       def failures
         test_results.select(&:fail?)
       end
+
+      def reset
+        [
+          'required_passed',
+          'required_total',
+          'error_count',
+          'skip_count',
+          'optional_passed',
+          'optional_total'
+        ].each { |field| send("#{field}=", 0) }
+      end
     end
   end
 end
--- a/lib/app/models/sequence_result.rb
+++ b/lib/app/models/sequence_result.rb
@@ -59,7 +59,7 @@ module Inferno
         test_results.select(&:fail?)
       end
 
-      def reset
+      def reset!
         [
           'required_passed',
           'required_total',
--- a/lib/app/models/sequence_result.rb
+++ b/lib/app/models/sequence_result.rb
@@ -34,6 +34,30 @@ module Inferno
 
       has n, :test_results, order: [:test_index.asc]
       belongs_to :testing_instance
+
+      def fail?
+        result == 'fail' || error?
+      end
+
+      def error?
+        result == 'error'
+      end
+
+      def pass?
+        result == 'pass'
+      end
+
+      def skip?
+        result == 'skip'
+      end
+
+      def wait?
+        result == 'wait'
+      end
+
+      def failures
+        test_results.select(&:fail?)
+      end
     end
   end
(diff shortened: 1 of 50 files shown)


lib/app/utils/web_driver.rb
```diff
@@ -29,13 +29,11 @@ def run_script(json_script, start_url = nil)
...
         unless command['find_value'].nil?
           current_element = wait.until do
-            current = if !command['index'].nil?
-                        driver.find_elements(command['type'].to_sym => command['find_value'])
-                      else
-                        driver.find_element(command['type'].to_sym => command['find_value'])
-                      end
-
-            current if (current.is_a?(Array) && current.length >= (command['index'] || 0) && current[command['index']].displayed?) || (!current.is_a?(Array) && current.displayed?)
+            if !command['index'].nil?
```
radamson: Could replace this if..else..end with

```
if command['index'].nil?
  driver.find_element(command['type'].to_sym => command['find_value'])
else
  driver.find_elements(command['type'].to_sym => command['find_value'])
end
```

to get rid of the `!`

lib/app/models/sequence_result.rb
```diff
@@ -34,6 +34,41 @@ class SequenceResult
...
+
+      def wait?
+        result == 'wait'
+      end
+
+      def failures
+        test_results.select(&:fail?)
+      end
+
+      def reset
```
arscan: Should this be `reset!`?

lib/app/models/testing_instance.rb
```diff
@@ -74,22 +74,20 @@ def group_results(test_set_id)
         results = latest_results_by_case
 
         self.module.test_sets[test_set_id.to_sym].groups.each do |group|
-          result_details = group.test_cases.each_with_object(cancel: 0, pass: 0, skip: 0, fail: 0, error: 0, total: 0) do |val, hash|
-            next unless results.key?(val.id)
+          result_details = group.test_cases.each_with_object(Hash.new(0)) do |test_case, hash|
```
arscan: `Hash.new(0)`, good to know!

lib/app/models/testing_instance.rb
```diff
@@ -99,38 +97,39 @@ def waiting_on_sequence
...
 
       def patient_id=(patient_id)
         return if patient_id.to_s == self.patient_id.to_s
 
-        resource_references.select { |ref| ref.resource_type == 'Patient' }
-        # Use destroy directly (instead of on each, so we don't have to reload)
-        resource_references.destroy
+        resource_references
+          .select { |ref| ref.resource_type == 'Patient' }
+          .destroy # Use destroy directly (instead of on each, so we don't have to reload)
```
arscan: Hmmm, was this a bug before, or did `select` mutate `resource_references`?  Seems like a bug.

lib/app/models/sequence_result.rb
Jammjammjamm: Sure.

lib/app/models/testing_instance.rb
Jammjammjamm: Oops, I didn't notice that I changed the behavior here. Do you know whether the intent is to delete all resources when the patient changes?

PR #208, opened 2019-06-04
Update capability statement sequence and remove US Core R4 from seque…

czh-orz:
Changed all references to "conformance" to "capability" in capability test. Also removed "UsCoreR4 from sequence titles"


lib/app/modules/us_core_r4/pediatric_bmi_for_age_sequence.rb
```diff
@@ -14,236 +14,253 @@ class PediatricBmiForAgeSequence < SequenceBase
...
+          
         when 'patient'
-          assert (resource&.subject && resource.subject.reference.include?(value)), 'patient on resource does not match patient requested'
-
+          assert (resource&.subject && resource.subject.reference.include?(value)), "patient on resource does not match patient requested"
+      
         when 'status'
-          assert !resource&.status.nil? && resource&.status == value, 'status on resource did not match status requested'
-
+          assert resource&.status != nil && resource&.status == value, "status on resource did not match status requested"
```
Jammjammjamm: In this statement: 
```
resource&.status != nil && resource&.status == value
```
`resource&.status != nil` is redundant unless you want this assertion to fail in cases when `value` is `nil`. The same is also done in `any?` blocks, like on lines 30 & 35.

arscan:
Unfortunately I cannot merge this until we synchronize the style fixes that @jammjammjamm made in #202 and #204 into these generation scripts.

PR #211, opened 2019-06-10
Try Me Base URL

nicole-kennedy:
Adding a "domain" field to the presets in the config file will cause that preset to only appear if the base url of the current url is the same as the domain in the config file. If there is no domain specified in the config file, then the preset will always appear. This allows the presets to be configured for both localhost and for inferno.healthit.gov. 

These are the two base urls: 
https://inferno.healthit.gov/inferno/ (this one should be double checked)
http://localhost:4567

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -58,6 +58,14 @@ presets:
     client_id: vkPKDPcTIEMaw5Uf-DdUUtNMFMZaX0
     client_secret: LS1nY3JFU3FDeEs0cWoxQWF6TVJFNU05RmZZNGZhZ2Vwb2JYWjdSWWJGakwwNTZ2Vng=
     instructions: https://github.com/siteadmin/inferno/wiki/SITE-Preset-Instructions
+  site_test_healthit_gov:
+    name: SITE FHIR Sandbox
+    uri: https://fhir.sitenv.org/secure/fhir
+    module: onc
+    inferno_uri: https://infernotest.healthit.gov
+    client_id: TrToU5piE-dJ1g6PBG1elFV4r9KLmH
+    client_secret: ckFqb1ZhbmFMQS13WDE0c1dTLWxSdGRLSE8yUXpWNS1vSnd6azNrMmU3Y0JPdDRja3U=
+    instructions: https://github.com/siteadmin/inferno/wiki/SITE-Preset-Instructions
   site_local: 
     name: SITE FHIR Sandbox
     uri: https://fhir.sitenv.org/secure/fhir
--- a/config.yml
+++ b/config.yml
@@ -48,12 +48,21 @@ modules:
   - us_core_r4
   - argonaut
 
-# preset fhir servers: optional. Minimally requires name, uri, module, optional client_id, client_secret, scopes, instructions link
+# preset fhir servers: optional. Minimally requires name, uri, module, optional domain, client_id, client_secret, scopes, instructions link
 presets:
   site_healthit_gov: 
     name: SITE FHIR Sandbox
     uri: https://fhir.sitenv.org/secure/fhir
     module: onc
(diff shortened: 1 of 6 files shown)


config.yml
```diff
@@ -48,12 +48,21 @@ modules:
...
     module: onc
+    domain: https://inferno.healthit.gov/inferno/
     client_id: vkPKDPcTIEMaw5Uf-DdUUtNMFMZaX0
     client_secret: LS1nY3JFU3FDeEs0cWoxQWF6TVJFNU05RmZZNGZhZ2Vwb2JYWjdSWWJGakwwNTZ2Vng=
+    instructions: https://github.com/siteadmin/inferno/wiki/SITE-Preset-Instructions
+  site_local: 
+    name: SITE FHIR Sandbox
+    uri: https://fhir.sitenv.org/secure/fhir
+    module: onc
+    domain: http://localhost:4567
```
arscan: Could we change this to `inferno_uri`.  I think `domain` is a bit of an overloaded term -- I would expect 'domain' to include just the domain (inferno.healthit.gov), but we are including the full path of an inferno installation instead.

nicole-kennedy: Done!

lib/app/endpoint.rb
```diff
@@ -51,6 +51,14 @@ class Endpoint < Sinatra::Base
       set :static, true
       set :views, File.expand_path('views', __dir__)
       set(:prefix) { '/' << name[/[^:]+$/].underscore }
+
+      def render_index
+        unless defined?(settings.presets).nil? || settings.presets.nil? # rubocop:disable Style/IfUnlessModifier
+          presets = settings.presets.select { |_, v| v['inferno_uri'].nil? || v['inferno_uri'] == request.base_url }
```
arscan: I don't think this would match `https://inferno.healthit.gov/inferno`, as `request.base_url` is probably only `http://domain`, without path information.  One complication is that usually inferno can either be accessed at `http://domain/inferno` or `http://domain`, but that isn't the case on `healthit.gov` because the root is intended to show the community vs. program landing page.  And we have `http://domain/community` for the community instance.

I would recommend:
* look for exact match to the domain (base_url) here
* but also accept `v['inferno_url'] == request.base_url + Inferno::BASE_PATH`

Also, since we are dealing with user input in the config, perhaps we should get rid of any trailing slashes just in case, on both BASE_PATH and v['inferno_url'].  Otherwise people may put in `http://localhost/inferno/` but that doesn't match `http://localhost/inferno`, but logically they are the same.  Look into `chomp('/')`.

nicole-kennedy: Updated cases for matching

lib/app/endpoint.rb
```diff
@@ -51,6 +51,20 @@ class Endpoint < Sinatra::Base
       set :static, true
       set :views, File.expand_path('views', __dir__)
       set(:prefix) { '/' << name[/[^:]+$/].underscore }
+
+      def render_index
+        unless defined?(settings.presets).nil? || settings.presets.nil?
+          base_url = request.base_url
+          base_path = Inferno::BASE_PATH.chomp('/') unless Inferno::BASE_PATH.nil?
```
Jammjammjamm:
```suggestion
          base_path = Inferno::BASE_PATH&.chomp('/')
```

lib/app/endpoint.rb
```diff
@@ -51,6 +51,20 @@ class Endpoint < Sinatra::Base
...
+
+          presets = settings.presets.select do |_, v|
+            inferno_uri = v['inferno_uri'].chomp('/') unless v['inferno_uri'].nil?
```
Jammjammjamm:
```suggestion
            inferno_uri = v['inferno_uri']&.chomp('/')
```

PR #212, opened 2019-06-11
Add Continue button for guided tests 

nicole-kennedy:
Added a "Continue" and "Back" button to the guided tests. The buttons appear after the test is run, no matter the result. The buttons have the same functionality as clicking the tabs at the top. 

Screen Shot 2019-06-17 at 2 12 34 PM

### DIFF
--- a/lib/app/views/guided.erb
+++ b/lib/app/views/guided.erb
@@ -158,75 +158,69 @@
 
                   <% end %>
 
-                  <% case group_result.to_s
+                  <div class="group-message">
+                    <% case group_result.to_s
                       when 'pass' %>
-                      <div class="group-message">
-                        Result: Pass. Tests have successfully passed.
-                      </div>
-                      <%= markdown_to_html(group.overview) %>
-                      <button class='sequence-button btn btn-info btn-sm' data-group-id='<%=group.id%>'>
-                        <span class="oi oi-loop"></span> Rerun All Tests
-                      </button>
-                    <% when 'fail' %>
-                      <div class="group-message">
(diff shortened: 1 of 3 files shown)


arscan:
Looks good!  Could you also add a 'back' button for steps 2-4?

nicole-kennedy:
Updated with requested changes

lib/app/views/guided.erb
```diff
@@ -158,77 +158,92 @@
...
-                    <% when 'skip' %>
-                        <div class="group-message">
-                        Skipped
-                      </div>
-                      <%= markdown_to_html(group.overview) %>
-                      <% if group.run_skipped && group.test_cases.any?{ |test_case| sequence_results[test_case.id].try(:result) == 'skip'} %>
-                        <button class='sequence-button btn btn-info btn-sm' data-group-id='<%=group.id%>' data-skipped-only='true'>
-                          <span class="oi oi-media-play"></span> Rerun Skipped Tests
+                      <% when 'fail' %>
+                        <% if group.run_skipped && group.test_cases.any?{ |test_case| sequence_results[test_case.id].try(:result) == 'skip'} %>
```
Jammjammjamm: Assuming that `group` is a `TestGroup`, this condition looks like it could be a `TestGroup` instance method. Then this and the ifs could just be something like `if group.skipped?`

arscan: The group doesn't have the results attached to it, it is just the metadata that defines the groups.  But I'll clean this up a little.

PR #213, opened 2019-06-13
Progress bar

nicole-kennedy:
Added a progress bar for when tests are running. 

Screen Shot 2019-06-13 at 1 57 11 PM

### DIFF
--- a/lib/app/endpoint/home.rb
+++ b/lib/app/endpoint/home.rb
@@ -121,11 +121,7 @@ module Inferno
 
               submitted_test_cases_count = sequence_result.next_test_cases.split(',')
               total_tests = submitted_test_cases_count.reduce(first_test_count) do |total, set|
-                sequence_test_count =
-                  test_set
-                    .test_case_by_id(set)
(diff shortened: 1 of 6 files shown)


lib/app/endpoint/home.rb
```diff
@@ -385,6 +399,18 @@ class Home < Endpoint
...
           client.default_json
           submitted_test_cases = params[:test_case].split(',')
+
+          instance.reload # ensure that we have all the latest data
+
+          total_tests = submitted_test_cases.reduce(0) do |total, set|
+            sequence_test_count =
+              test_set
+                .test_case_by_id(set)
+                .sequence.new(instance, client, settings.disable_tls_tests)
```
arscan: I don't think you need to instantiate sequence here... `test_count` should work on the class as well.  Did you try that and it didn't work?

nicole-kennedy: You're right, didn't realize that wasn't needed. All updated now, thanks!

PR #216, opened 2019-06-17
EHR launch redirect

nicole-kennedy:
When launching an app from an external website when the Inferno launch test is not actively running, it will redirect back to Inferno with the previous session restored and an error message if there are cookies from a previous session. The cookie is updated when the instance is created and whenever a test sequence is executed.

Screen Shot 2019-06-17 at 4 16 07 PM

This also alters the behavior such that if an invalid state is passed during a redirect (which traditionally would result in an ambigious error page), Inferno figures out which tests are running and allows them to continue, but failing the 'state' test.  This enables much more consistent feedback for this type of error:

Screen Shot 2019-06-21 at 4 14 36 PM

If there are no cookies stored, then there is the same error page as before:  

Screen Shot 2019-06-17 at 4 16 24 PM


lib/app/views/guided.erb
```diff
@@ -69,6 +69,9 @@
         <% elsif error_code == 'no_redirect' %>
           The redirect URI for this instance was visited, but no sequence was actively running that was waiting for this event.  Please run the a launch sequence test
           before visiting this redirect url.
+        <% elsif error_code == 'ehr_launch' %>
+          No actively running launch sequences found for iss <%=instance.url%>.
```
arscan: This isn't quite correct.  The `iss` is the URI being passed as a query parameter from the auth server, which could be different than what we have in our instance (and hence the error).  You need to pass that along in the querystring  so you can grab it here.  For example, `error_code=ehr_launch&iss=params[:iss]`.  Though you may want to doublecheck that the iss that gets passed in the querystring is URI encoded/escaped properly.

nicole-kennedy: Think it's updated to what you wanted, fixed iss and added checking for waiting state

lib/app/sequence_base.rb
```diff
@@ -61,11 +61,16 @@ def initialize(instance, client, disable_tls_tests = false, sequence_result = ni
         @metadata_only = metadata_only
       end
 
-      def resume(request = nil, headers = nil, params = nil, &block)
+      def resume(request = nil, headers = nil, params = nil, fail_message = nil, &block)
         @params = params unless params.nil?
 
         @sequence_result.test_results.last.result = STATUS[:pass]
 
+        unless fail_message.nil?
```
Jammjammjamm:
```suggestion
        if fail_message.present?
```

lib/app/endpoint/home.rb
```diff
@@ -23,33 +24,33 @@ class Home < Endpoint
...
+              instance = Inferno::Models::TestingInstance.get(cookies[:instance_id_test_set]&.split('/')&.first)
+              if instance.nil?
+                error_message = '<p>Inferno has detected an issue with the SMART launch. ' \
+                                "No actively running launch sequences found with a state of #{params[:state]}. " \
+                                'The authorization server is not returning the correct state variable and ' \
+                                'therefore Inferno cannot identify which server is currently under test. ' \
+                                "Please click your browser's \"Back\" button to return to Inferno, " \
+                                "and click \"Refresh\" to ensure that the most recent test results are visible.</p>" # rubocop: disable Style/StringLiterals
+
+                error_message += "<p>Error returned by server: <strong>#{params[:error]}</strong>.</p>" unless params[:error].nil?
```
Jammjammjamm:
```suggestion
                error_message += "<p>Error returned by server: <strong>#{params[:error]}</strong>.</p>" if params[:error].present?
```

lib/app/endpoint/home.rb
```diff
@@ -23,33 +24,33 @@ class Home < Endpoint
...
-
-              halt 500, error_message
+              instance = Inferno::Models::TestingInstance.get(cookies[:instance_id_test_set]&.split('/')&.first)
+              if instance.nil?
+                error_message = '<p>Inferno has detected an issue with the SMART launch. ' \
+                                "No actively running launch sequences found with a state of #{params[:state]}. " \
+                                'The authorization server is not returning the correct state variable and ' \
+                                'therefore Inferno cannot identify which server is currently under test. ' \
+                                "Please click your browser's \"Back\" button to return to Inferno, " \
+                                "and click \"Refresh\" to ensure that the most recent test results are visible.</p>" # rubocop: disable Style/StringLiterals
```
Jammjammjamm: The rubocop disable only affects this line, right? This line could use single quotes and unescape the double quotes rather than ignoring the cop.

arscan: Yeah we shouldn't disable the cop unless there is a good reason. Any objection to me moving this to `%()` @Jammjammjamm ?

Jammjammjamm: Sounds good to me.

PR #218, opened 2019-06-20
Initial Must Support Implementation on US Core R4

czh-orz:
This creates tests for must-support elements. For elements with variable types (it contains [x]), it checks each possible value of [x].


generators/argonaut-r4/metadata_extractor.rb
```diff
@@ -119,6 +119,39 @@ def add_interactions(resource, sequence)
...
+            id: element['id'],
+            path: element['path'],
+            url: element['type'].first['profile'].first
+          }
+        next
+      end
+
+      path = element['path']
+      if path.include? '[x]'
+        variable_el = profile_definition['snapshot']['element'].select { |el| el['id'] == (path.split('[x]').first + '[x]') }.first
```
Jammjammjamm: Use `.find {}` instead of `.select {}.first`. Also, "choice elements" is the official term for these, so maybe `variable` should be replaced with `choice`.

lib/app/sequence_base.rb
```diff
@@ -23,14 +23,7 @@ class SequenceBase
...
 
-      STATUS = {
-        pass: 'pass',
-        fail: 'fail',
-        error: 'error',
-        todo: 'todo',
-        wait: 'wait',
-        skip: 'skip'
-      }.freeze
+      STATUS = { pass: 'pass', fail: 'fail', error: 'error', todo: 'todo', wait: 'wait', skip: 'skip' }.freeze
```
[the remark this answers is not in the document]
czh-orz: This is just for the linter. Four lines over 😢

generators/argonaut-r4/metadata_extractor.rb
czh-orz: Changed to find and used choice.

lib/app/sequence_base.rb
```diff
@@ -746,6 +739,17 @@ def check_resource_against_profile(resource, resource_type, specified_profile =
...
         assert(errors.empty?, errors.join("<br/>\n"))
       end
+
+      def can_resolve_path(element, path)
+        return true if path.empty?
+
+        path_ary = path.split('.')
+        el_as_array = Array.wrap(element)
+        cur_path_part = path_ary.delete_at(0)
+        return false unless el_as_array.any? { |el| !el.try(cur_path_part.to_sym).nil? }
```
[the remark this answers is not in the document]
czh-orz: changed to suggestion

lib/app/sequence_base.rb
```diff
@@ -746,6 +739,17 @@ def check_resource_against_profile(resource, resource_type, specified_profile =
         end
         assert(errors.empty?, errors.join("<br/>\n"))
       end
+
+      def can_resolve_path(element, path)
+        return true if path.empty?
+
+        path_ary = path.split('.')
+        el_as_array = Array.wrap(element)
+        cur_path_part = path_ary.delete_at(0)
```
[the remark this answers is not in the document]
czh-orz: Updated to suggestion

lib/app/modules/us_core_r4/pediatric_bmi_for_age_sequence.rb
```diff
@@ -217,9 +217,105 @@ def validate_resource_item(resource, property, value)
...
+      test 'Demonstrates that the server can supply must supported elements' do
         metadata do
           id '10'
+          link 'https://build.fhir.org/ig/HL7/US-Core-R4/general-guidance.html/#must-support'
+          desc %(
+          )
+          versions :r4
+        end
+
+        element_found = @instance.must_support_confirmed.include?('Observation.status') || can_resolve_path(@observation, 'status')
```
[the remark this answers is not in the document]
czh-orz: Addressed. Updated to loop through array of elements/extensions

generators/argonaut-r4/generator.rb
```diff
@@ -139,6 +140,37 @@ def create_interaction_test(sequence, interaction)
   sequence[:tests] << interaction_test
 end
 
+def create_must_support_test(sequence)
+  test = {
+    tests_that: 'Demonstrates that the server can supply must supported elements',
+    index: format('%02d', sequence[:tests].length + 1),
```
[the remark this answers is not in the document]
czh-orz: It's because it needs to be like '02' instead of 2, but I just updated it to be formatted it in the erb file instead.

lib/app/modules/us_core_r4/pediatric_bmi_for_age_sequence.rb
```diff
@@ -217,9 +217,58 @@ def validate_resource_item(resource, property, value)
...
           id '10'
+          link 'https://build.fhir.org/ig/HL7/US-Core-R4/general-guidance.html/#must-support'
+          desc %(
+          )
+          versions :r4
+        end
+
+        must_support_elements = [
+          'Observation.status',
+          'Observation.category',
```
arscan: does it matter that `Observation.category` is listed twice?

arscan: Ah, probably has to do with the `category` slice.  Did you explicitly take into account slices or is that a potential gap?

generators/argonaut-r4/generator.rb
```diff
@@ -139,10 +140,61 @@ def create_interaction_test(sequence, interaction)
...
+
+  test[:test_code] += %(
+        @instance.save!)
+
+  sequence[:tests] << test
+end
+
 def create_resource_profile_test(sequence)
   test = {
     tests_that: "#{sequence[:resource]} resources associated with Patient conform to Argonaut profiles",
```
arscan: Could you rename to 'US Core R4 profiles'?

arscan:
Could you also rename the `argonaut-r4` directory to `uscore-r4`?

lib/app/modules/us_core_r4/us_core_device_sequence.rb
```diff
@@ -145,9 +145,35 @@ def validate_resource_item(resource, property, value)
...
+          'Device.udiCarrier',
+          'Device.udiCarrier.carrierAIDC',
+          'Device.udiCarrier.carrierHRF',
+          'Device.type',
+          'Device.patient'
+        ]
+        must_support_elements.each do |path|
+          truncated_path = path.gsub('Device.', '')
+          already_found = @instance.must_support_confirmed.include?(path)
+          element_found = already_found || can_resolve_path(@device, truncated_path)
```
arscan: It appears that we are just checking a single device, even if they returned multiple devices from the searches.  I think we should look across all devices provided for the patient, and pass them if at least one has a must support element.

lib/app/modules/us_core_r4/pediatric_bmi_for_age_sequence.rb
czh-orz: It's a gap. I didn't know how to handle slices so I just left them there for now.

generators/argonaut-r4/generator.rb
czh-orz: Done

lib/app/modules/us_core_r4/us_core_device_sequence.rb
czh-orz: I made it check all the resources in the reply. We should probably make it check all the resources for the other tests as well... but I think that should be another task.

lib/app/modules/us_core_r4/us_core_allergyintolerance_sequence.rb
```diff
@@ -161,16 +162,18 @@ def validate_resource_item(resource, property, value)
...
-          truncated_path = path.gsub('AllergyIntolerance.', '')
-          already_found = @instance.must_support_confirmed.include?(path)
-          element_found = already_found || can_resolve_path(@allergyintolerance, truncated_path)
-          skip "Could not find #{path} in the provided resource" unless element_found
-          @instance.must_support_confirmed += "#{path}," unless already_found
+          @allergyintolerance_ary.each do |resource|
+            truncated_path = path.gsub('AllergyIntolerance.', '')
+            already_found = @instance.must_support_confirmed.include?(path)
+            element_found = already_found || can_resolve_path(resource, truncated_path)
+            skip "Could not find #{path} in the provided resource" unless element_found
```
arscan: Skip will cause the whole test to skip on the first resource, if the first resource doesn't have all the elements, right?  The idea is that it should skip if by the end of looking at every resource I never saw at least one element.

lib/app/modules/us_core_r4/us_core_device_sequence.rb
arscan: What do you mean by 'the other tests'?

lib/app/modules/us_core_r4/us_core_allergyintolerance_sequence.rb
czh-orz: should be fixed now

PR #220, opened 2019-06-24
Update button border color

nicole-kennedy:
Updated border color on Next/Back/Run buttons to match their background color.


arscan:
Good catch :+1:

PR #222, opened 2019-06-25
Fix report rollup

arscan:
The rollup report didn't work properly because it required all sequences stored in inferno to have been passed.  Since we added the concept of modules and test_sets, this doesn't work anymore.  Updated to use the new test structures.

Also, before it would mark as 'fail', even if you hadn't run any tests yet.  Updated to mark as 'incomplete' if you hadn't passed everything but haven't failed anything either.

Screen Shot 2019-06-24 at 9 50 58 PM

### DIFF
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -98,7 +98,7 @@ module Inferno
       end
 
       def all_test_cases(test_set_id)
-        self.module.test_sets[test_set_id.to_sym].groups.each_with_object([]) { |g, o| o.push(*g.test_cases) }
+        self.module.test_sets[test_set_id.to_sym].groups.flat_map(&:test_cases)
       end
 
       def all_passed?(test_set_id)
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -2,6 +2,7 @@
 
 require 'dm-core'
 require 'dm-migrations'
+require_relative '../utils/result_statuses'
 
 module Inferno
   module Models
@@ -119,9 +120,9 @@ module Inferno
 
       def final_result(test_set_id)
         if all_passed?(test_set_id)
-          'pass'
+          Inferno::ResultStatuses::PASS
         else
-          any_failed?(test_set_id) ? 'fail' : 'incomplete'
+          any_failed?(test_set_id) ? Inferno::ResultStatuses::FAIL : Inferno::ResultStatuses::PENDING
         end
       end
 
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -97,16 +97,32 @@ module Inferno
         sequence_results.first(result: 'wait')
       end
 
-      def all_passed?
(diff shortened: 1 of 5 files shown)


lib/app/models/testing_instance.rb
```diff
@@ -97,16 +97,32 @@ def waiting_on_sequence
...
+        self.module.test_sets[test_set_id.to_sym].groups.each_with_object([]) { |g, o| o.push(*g.test_cases) }
+      end
+
+      def all_passed?(test_set_id)
+        latest_results = latest_results_by_case
 
-        required_sequences.all? do |sequence|
-          latest_results[sequence.name].try(:result) == 'pass'
+        all_test_cases(test_set_id).all? do |test_case|
+          latest_results[test_case.id]&.result == 'pass'
```
Jammjammjamm:
```suggestion
          latest_results[test_case.id]&.pass?
```

lib/app/models/testing_instance.rb
```diff
@@ -97,16 +97,32 @@ def waiting_on_sequence
...
         end
       end
 
-      def final_result
-        all_passed? ? 'pass' : 'fail'
+      def any_failed?(test_set_id)
+        latest_results = latest_results_by_case
+
+        all_test_cases(test_set_id).any? do |test_case|
+          ['fail', 'error'].include? latest_results[test_case.id]&.result
```
Jammjammjamm:
```suggestion
         latest_results[test_case.id]&.fail?
```
`#fail?` is true if there is a failure or an error.

lib/app/models/testing_instance.rb
```diff
@@ -97,16 +97,32 @@ def waiting_on_sequence
         sequence_results.first(result: 'wait')
       end
 
-      def all_passed?
-        required_sequences = Inferno::Sequence::SequenceBase.subclasses.reject(&:optional?)
+      def all_test_cases(test_set_id)
+        self.module.test_sets[test_set_id.to_sym].groups.each_with_object([]) { |g, o| o.push(*g.test_cases) }
```
Jammjammjamm:
```suggestion
        self.module.test_sets[test_set_id.to_sym].groups.flat_map(&:test_cases)
```
I think the above should be equivalent.

lib/app/models/testing_instance.rb
```diff
@@ -97,16 +97,32 @@ def waiting_on_sequence
...
+        all_test_cases(test_set_id).any? do |test_case|
+          ['fail', 'error'].include? latest_results[test_case.id]&.result
+        end
+      end
+
+      def final_result(test_set_id)
+        if all_passed?(test_set_id)
+          'pass'
+        else
+          any_failed?(test_set_id) ? 'fail' : 'incomplete'
```
Jammjammjamm: I think we should use the constants defined in `Inferno::ResultStatuses` for these strings. One will have to be added for `incomplete`.

arscan: I move to constants, and renamed to `pending`.  Ideally I would have just mixed in ResultStatuses, but due to our current structure we couldn't just refer to a `results` method because we need to know which 'test set' we are in.  Hopefully this is good enough for now and we can keep this in mind when working on future designs.

PR #224, opened 2019-06-25
Fix DSTU2 CareTeam profile validation

Jammjammjamm:
This branch modifies how resources are validated against a profile. Previously, passing a profile to `SequenceBase#test_resources_against_profile` didn't guarantee that the resources were actually validated against the specified profile. Now, when it receives a specific profile, it:

- Validates any resources which were saved and categorized as conforming to that profile
- If no resources have been saved with that profile, it checks `meta.profile` for resources which claim to conform to the profile and validates those
- If no resources were found, the test is skipped

I did a fair amount of refactoring while investigating this problem, so the actual bug fix is all in [this commit.](https://github.com/siteadmin/inferno/pull/224/commits/3fb4ba8383d8b8e2c77133512ca5a0433b46898a)

**NOTE:** the `sequence_base.rb` diff is collapsed by default, so be sure to expand it

### DIFF
--- a/lib/app/endpoint/home.rb
+++ b/lib/app/endpoint/home.rb
@@ -129,7 +129,7 @@ module Inferno
 
               out << js_hide_wait_modal
               out << js_show_test_modal
-              count = sequence_result.test_results.length
+              count = sequence_result.result_count
 
               submitted_test_cases_count = sequence_result.next_test_cases.split(',')
               total_tests = submitted_test_cases_count.reduce(first_test_count) do |total, set|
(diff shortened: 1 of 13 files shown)


arscan:
Shouldn't the ArgonautCarePlanSequence specify the careplan profile when saving the references?  It doesn't do that now, which is causing the sequence to skip on the SITE preset.  But if you look what SITE sends back, it looks like it is the correct profile.  

I didn't check all of tests, but in these tests all resources must conform to one argo profile or another (determined by what query call they came from).

Jammjammjamm:
I think that all the tests should specify a profile to test against unless they are just testing base resources. I'll look into it.

PR #226, opened 2019-06-26
Results progress bar

nicole-kennedy:
Progress bar updates in real time with results corresponding to the results icon colors. 

Screen Shot 2019-06-25 at 9 24 38 PM
Screen Shot 2019-06-25 at 9 24 59 PM


lib/app/endpoint/home.rb
```diff
@@ -129,17 +133,14 @@ class Home < Endpoint
...
-              end
 
+              prev_count = sequence_result.test_results.length
+              count = prev_count
               sequence_result = sequence.resume(request, headers, request.params, error_message) do |result|
+                rerun = count > prev_count ? nil : instance.previous_test_results(previous_test_sequences).to_a + sequence_result.test_results.map(&:result).to_a
                 count += 1
-                out << js_update_result(sequence, test_set, result, count, sequence.test_count, count, total_tests)
+                test_count += 1
+                out << js_update_result(sequence, test_set, result, count, sequence.test_count, test_count, total_tests, rerun)
```
arscan: Instead of altering the js_update_result method to optionally print out previous test score, could you just loop through the previous tests and call js_update_result (the old version) on each here?

nicole-kennedy: Yes it could but then the other information passed into `js_update_result` like `sequence, _test_set, result, set_count, set_total, count` would technically be incorrect and the `console.log` would get repeated for the "rerun" tests. As long as that's okay, do you still want me to update it?

arscan: Yeah I guess my mental model was that it would just replay what already happened, so the user would be flashed very briefly everything that already run, but it would make it so you had one less special case in the code.  But perhaps your method just makes more sense.  I think you can just leave it.

nicole-kennedy: That makes sense too and it's an easy fix if you want to, I had just originally done it like that because `js_update_result` wouldn't know that some of the parameters were inaccurate but it wouldn't actually change anything for the user (only the console log).

PR #227, opened 2019-06-26
Remove DB from MUST SUPPORT checks

arscan:
@czh-orz -- I made a couple of potential changes.  The main difference is that I don't think we should save any data in TestingInstance, because we don't want repeated runs of the same sequence (with same inputs, outputs, server response) to return different results.  I think just requiring that we receive all MUST SUPPORT elements across all resource_x associated with a single patient is good enough.

I was unable to run the code due to issues retrieving the profiles, so don't just merge this (there are likely bugs).  But are you ok with this approach?


arscan:
FYI @czh-orz -- I'm going to merge in these updates into your MUST SUPPORT branch because it keeps things a little simpler by relying on the DB to track.  It may require a rebase for your later PRs.  We can discuss if there is an issue.

PR #231, opened 2019-06-27
Test cleanup

Jammjammjamm:
This branch:

- Removes all non-error output from tests. Test output is logged to `tmp/test.log`
- Replaces all uses of `FHIR.logger` with `Inferno.logger`
- Makes the csv generating rake task write files to `tmp` during tests so that git ignores them.
- Fixes a minitest deprecation warning
- Adds `pry-byebug` so you can step through code line by line while using pry


arscan:
I am very surprised your fix in https://github.com/siteadmin/inferno/pull/231/commits/ec85c33d7defd14a82ee7f585a43b5f41331fdf8 worked.  I could try to track down why this didn't work for us earlier (I believe we were waiting on some other dependency, which might be resolved by now).  But if it works then we are fine.

Could you check https://github.com/siteadmin/inferno-site-overlay to make sure that this doesn't affect the way we are packaging and running inferno.healthit.gov?  I doubt it would, but its worth checking.

arscan:
While you are looking at this, it might be worth streamlining our `docker` tests somehow.  I'm not sure if it is valuable to rerun all ruby tests, as that seems redundant.   I added that to ensure that we maintained the docker config and it didn't break, but waiting an extra 3+ minutes (and growing) to redo the full barrage of unit tests seems like a waste of electricity.

PR #232, opened 2019-07-01
Fi 241 check correct version

hershilpatel:
Branch that added a new test to capability_statement_sequence.rb to confirm that the client version of fhir matches with the selected version.


arscan:
Could you point the merge target of this PR to the development branch?  We use that branch as a staging area prior to deploying an official new version.

lib/app/modules/core/capability_statement_sequence.rb
```diff
@@ -71,9 +71,22 @@ class CapabilityStatementSequence < SequenceBase
         end
       end
 
-      test 'FHIR server supports the conformance interaction that defines how it supports resources' do
+      test 'Test FHIR version' do
```
radamson: Let's try and be more explicit about what `Test FHIR version` really means.
```suggestion
      test 'FHIR version of the server matches the FHIR version expected by tests' do
```

lib/app/modules/core/capability_statement_sequence.rb
```diff
@@ -71,9 +71,22 @@ class CapabilityStatementSequence < SequenceBase
...
         metadata do
           id '02'
+          link 'http://www.hl7.org/fhir/directory.cfml'
+          desc %(
+            Testing that the FHIR client is the same version that was version as the testing instance that was selected.
```
radamson: The existing sentence doesn't really make sense.  Can we make this a little clearer and detailed?
```suggestion
            Checks that the FHIR version of the server matches the FHIR version expected by the tests.
            This test will inspect the CapabilityStatement returned by the server to verify the FHIR version of the server.
```

Gemfile.lock
```diff
@@ -253,4 +253,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   2.0.1
+   1.16.1
```
Jammjammjamm: `gem install bundler && bundle` to get rid of this change.

lib/app/modules/core/capability_statement_sequence.rb
```diff
@@ -71,9 +71,23 @@ class CapabilityStatementSequence < SequenceBase
...
+      test 'FHIR version of the server matches the FHIR version expected by tests' do
         metadata do
           id '02'
+          link 'http://www.hl7.org/fhir/directory.cfml'
+          desc %(
+            Checks that the FHIR version of the server matches the FHIR version expected by the tests.
+            This test will inspect the CapabilityStatement returned by the server to verify the FHIR version of the server.
+          )
+        end
+
```
arscan: Technically @instance.fhir_version can be `nil`, which means that the tests you are running are version independant.  This is unusual, but is the case for our SMART on FHIR launch tests technically.  Just add a line like this to automatically have them pass in this case:
```
pass 'Tests are not version dependent' if @instance.fhir_version.blank?
```

Having a test `pass` doesn't make a ton of sense, but it is the closest status we have up to this point.  A ticket for you to address in the near future would be to add an `omit` status, which would be perfect for these types of cases.

Gemfile.lock
hershilpatel: Ran the gem install bundler && bundle, and it updated to 2.0.2 (which I assume is fine?)

lib/app/modules/core/capability_statement_sequence.rb
hershilpatel: Pushed new version with this fix

lib/app/modules/core/capability_statement_sequence.rb
```diff
@@ -71,9 +71,25 @@ class CapabilityStatementSequence < SequenceBase
...
+        pass 'Tests are not version dependent' if @instance.fhir_version.blank?
+
+        # @client.detect_version is a symbol
+        assert_equal(@instance.fhir_version.upcase, @client.detect_version.to_s.upcase, 'FHIR client version does not match with instance version')
```
arscan:
```suggestion
        assert_equal(@instance.fhir_version.upcase, @client.detect_version.to_s.upcase, 'FHIR client version does not match with instance version.')
```

hershilpatel:
That makes sense. Perhaps we should display the id number based on the order it appears in the list of tests in the sequence rather than explicitly supplying an id? In either case, I’ll update the extended sequences.

Thanks,

Hershil

hershilpatel:
That makes sense. Perhaps we should display the id number based on the order it appears in the list of tests in the sequence rather than explicitly supplying an id? Or do we need an explicit id for documentation purposes? In either case, I’ll update the extended sequences.

arscan:
The idea is that we would like to provide unique IDs that do not change over time, so if you fail TR-03 today, it will be the same as TR-03 in four years, even if we add or delete tests.  We could name this test id "02a" or something (like you would do with highway exits if one is added), but since ONC hasn't yet finalized program certification criteria, I think it is premature to lock in the test numbers.  Might as well keep it clean.

PR #234, opened 2019-07-02
FI-205: Add ValueSets for R4 Terminology Validation

radamson:
Add additional ValueSets and CodeSystems for US Core STU3 (FHIR R4) terminology validation and registers the validators with `FHIR::Models`.

Also removed some of the previous terminology code that is no longer used.

Screen Shot 2019-07-01 at 2 19 41 PM


Jammjammjamm:
How do we verify that this works?

radamson:
You can exercise it manually by running it against an R4 HSPC Sandbox.  Adams, Daniel (SMART-1288992) has some MedicationRequests that trigger warnings as shown.

I'll add a test too for `SequenceBase#validate_resource` that uses this as well.  If you think it would be better to test it elsewhere, let me know.

Jammjammjamm:
Let's talk in person about this, I don't see those errors.

arscan:
Will this work for both US Core v.3.0.0 and US Core v.3.0.1?  Or are there assumptions about what won't change in US Core going forward?

PR #241, opened 2019-07-11
Fi 258 fail on launches if no auth

hershilpatel:
Fixed issue where on redirect, if the returned oauth authorization uri is empty it will redirect to itself (also handle cases of invalid oauth authorization uri in general)

Added new assert_valid_http_uri 

Updated standalone_launch_sequence 
   02 to assert that the oauth authorization endpoint is valid before attempting to redirect 
   updated the other propagating tests to not crash when 02 fails to return a valid params and a valid tokens

Updated smart_discovery_sequence.rb
   replaced uri checkes with the new assert_valid_http_uri

### DIFF
--- a/lib/app/modules/smart/standalone_launch_sequence.rb
+++ b/lib/app/modules/smart/standalone_launch_sequence.rb
@@ -35,8 +35,8 @@ module Inferno
         !@instance.client_id.nil?
       end
 
-      OAUTH_REDIRECT_FAILED = "Redirect to OAuth server failed"
-      NO_TOKEN = "No valid token"
+      OAUTH_REDIRECT_FAILED = 'Redirect to OAuth server failed'
+      NO_TOKEN = 'No valid token'
 
       test 'OAuth 2.0 authorize endpoint secured by transport layer security' do
         metadata do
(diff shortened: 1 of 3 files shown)


lib/app/modules/smart/standalone_launch_sequence.rb
```diff
@@ -96,6 +104,9 @@ class StandaloneLaunchSequence < SequenceBase
           )
         end
 
+        # Confirm that there is a @params object from the redirect
+        assert !@params.nil?, OAUTH_REDIRECT_FAILED
```
Jammjammjamm: Should this assertion pass if `@params` is an empty object?

lib/app/utils/assertions.rb
```diff
@@ -260,5 +262,13 @@ def assert_deny_previous_tls(uri)
           )
       end
     end
+
+    def assert_valid_http_uri(uri, message = nil)
+      error_message = message
+      if error_message.nil?
```
Jammjammjamm: `error_message = message || "\"#{uri}\" is not a valid URI"`

lib/app/modules/smart/standalone_launch_sequence.rb
hershilpatel: I believe the future asserts will fail (with a more accurate error message), indicating desired fields don't exist in the hash

lib/app/utils/assertions.rb
hershilpatel: change incorporated and pushed

PR #242, opened 2019-07-11
FI-271 clinical notes

yunwwang:
Created sequence for ClincalNotes, two text fixtures and text text files for the sequence

### DIFF
--- a/lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
+++ b/lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
@@ -175,7 +175,7 @@ module Inferno
       test 'Server returns Pathology report from DiagnosticReport search by patient+category' do
         metadata do
           id '08'
-          link 'https://build.fhir.org/ig/HL7/US-Core-R4/clinical-notes-guidance.html'
+          link 'https://www.hl7.org/fhir/us/core/clinical-notes-guidance.html'
           desc %(
           )
           versions :r4
--- a/lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
+++ b/lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
@@ -115,7 +115,7 @@ module Inferno
       test 'Server shall have Discharge Summary' do
         metadata do
           id '03'
-          link 'https://build.fhir.org/ig/HL7/US-Core-R4/clinical-notes-guidance.html'
+          link 'https://www.hl7.org/fhir/us/core/clinical-notes-guidance.html'
           desc %(
           )
           versions :r4
--- a/lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
+++ b/lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
@@ -1,0 +1,232 @@
+# frozen_string_literal: true
+
+module Inferno
+  module Sequence
+    class UsCoreR4ClinicalNotesSequence < SequenceBase
+      group 'US Core R4 Profile Conformance'
+
+      title 'Clinical Notes Guideline Tests'
+
+      description 'Verify that DocumentReference and DiagnosticReport resources on the FHIR server follow the US Core R4 Clinical Notes Guideline'
+
+      test_id_prefix 'ClinicalNotes' # change me
+
+      requires :token, :patient_id
+      conformance_supports :DocumentReference
+
+
+      details %(
+
+        The #{title} Sequence tests `#{title.gsub(/\s+/, '')}` resources associated with the provided patient.  The resources
+        returned will be checked for consistency against the [Documentreference Argonaut Profile](https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-documentreference)
+
+      )
+
+      @clinicalnotes_found = false
+
+      def test_clinicalnotes_documentreference(type_code)
+        skip 'No Clinical Notes appear to be available for this patient. Please use patients with more information.' unless @clinicalnotes_found
+
+        assert @must_have_type_code[type_code], "Clinical Notes shall have at least one DocumentReference with type #{type_code}"
+      end
+
+      def test_clinicalnotes_diagnosticreport(category_code)
+        skip 'No Clinical Notes appear to be available for this patient. Please use patients with more information.' unless @clinicalnotes_found
+
+        patient_val = @instance.patient_id
+        search_params = { 'patient': patient_val, 'category': category_code }
+
+        reply = get_resource_by_params(versioned_resource_class('DiagnosticReport'), search_params)
+        assert_response_ok(reply)
+        assert_bundle_response(reply)
+
+        resource_count = reply&.resource&.entry&.length || 0
+        @resources_found = true if resource_count.positive?
+
+        assert @resources_found, "Clinical Notes shall have at least one DiagnosticReport with category #{category_code}"
+
+        diagnosticreport_ary = reply&.resource&.entry&.map { |entry| entry&.resource }
+
+        diagnosticreport_ary&.each do |resource|
+          id = resource&.id
+
+          resource&.presentedForm&.each do |attachment|
+            url_value = attachment&.url
+            next if url_value.nil?
+            
+            assert @attachment_url.key?(url_value), "Attachment #{url_value} referenced in DiagnosticReport/#{id} but not in any DocumentReference"
+            @attachment_url[url_value][:flag] = true
+          end 
+        end
+
+      end
+
+      test 'Server returns expected results from DocumentReference search by patient+clinicalnotes' do
+        metadata do
+          id '01'
+          link 'https://build.fhir.org/ig/HL7/US-Core-R4/clinical-notes-guidance.html'
+          desc %(
+          )
+          versions :r4
+        end
+
+        patient_val = @instance.patient_id
+        category_val = 'clinical-note'
+        search_params = { 'patient': patient_val, 'category': category_val }
+        @attachment_url = {}
+
+        reply = get_resource_by_params(versioned_resource_class('DocumentReference'), search_params)
+        assert_response_ok(reply)        
+        assert_bundle_response(reply)
+
+        resource_count = reply&.resource&.entry&.length || 0
+        @clinicalnotes_found = true if resource_count.positive?
+
+        skip 'No Clinical Notes appear to be available for this patient. Please use patients with more information.' unless @clinicalnotes_found
+
+        documentreference_ary = reply&.resource&.entry&.map { |entry| entry&.resource }
+
+        @must_have_type_code = {
+          "11488-4" => false, #Consultation Note
+          "18842-5" => false, #Dischard Summary
+          "34117-2" => false, #History and physical note
+          "28570-0" => false, #Procedure note
+          "11506-3" => false #Progress note
+        }
+
+        documentreference_ary&.each do |resource|
+          codings = resource&.type&.coding
+          next if codings.nil?
+
+          codings&.each do |coding|
+            code = coding&.code
+            system = coding&.system
+
+            next if system != 'http://loinc.org' or  !@must_have_type_code.key?(code)
+
+            @must_have_type_code[code] = true
+
+            resource&.content&.each do |a_content|
+              url_value = a_content&.attachment&.url
+              next if url_value.nil? or @attachment_url.key?(url_value)
+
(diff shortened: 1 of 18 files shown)


lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,228 @@
...
+module Inferno
+  module Sequence
+    class UsCoreR4ClinicalNotesSequence < SequenceBase
+      group 'US Core R4 Profile Conformance'
+
+      title 'Clinical Notes Guideline Tests'
+
+      description 'Verify that DocumentReference and DiagnosticReport resources on the FHIR server follow the US Core R4 Clinical Notes Guideline'
+
+      test_id_prefix 'ClinicalNotes' # change me
```
Jammjammjamm: Don't forget to change this.

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,228 @@
...
+        metadata do
+          id '01'
+          link 'https://build.fhir.org/ig/HL7/US-Core-R4/clinical-notes-guidance.html'
+          desc %(
+          )
+          versions :r4
+        end
+
+        patient_val = @instance.patient_id
+        category_val = 'clinical-note'
```
Jammjammjamm: `_val` doesn't really add any meaning to a variable name. How about `patient_id` and `category`?

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,228 @@
...
+        reply = get_resource_by_params(versioned_resource_class('DiagnosticReport'), search_params)
+        assert_response_ok(reply)
+        assert_bundle_response(reply)
+
+        resource_count = reply&.resource&.entry&.length || 0
+        @resources_found = true if resource_count.positive?
+
+        assert @resources_found, "Clinical Notes shall have at least one DiagnosticReport with category #{category_code}"
+
+        diagnosticreport_ary = reply&.resource&.entry&.map { |entry| entry&.resource }
```
Jammjammjamm: `diagnostic_reports`

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,228 @@
...
+        reply = get_resource_by_params(versioned_resource_class('DocumentReference'), search_params)
+        assert_response_ok(reply)
+        assert_bundle_response(reply)
+
+        resource_count = reply&.resource&.entry&.length || 0
+        @clinicalnotes_found = true if resource_count.positive?
+
+        skip 'No Clinical Notes appear to be available for this patient. Please use patients with more information.' unless @clinicalnotes_found
+
+        documentreference_ary = reply&.resource&.entry&.map { |entry| entry&.resource }
```
Jammjammjamm: `document_references`

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,228 @@
...
+
+        @must_have_type_code = {
```
Jammjammjamm: I don't think a `Hash` is the right data type here. If all we want to do is check inclusion in a group, I think a `Set` would be more appropriate. We have a set of the codes we expect to see, and a set of codes we have seen. We compare the two sets to determine whether all codes have been seen.

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,228 @@
...
+          '11488-4' => false, # Consultation Note
+          '18842-5' => false, # Dischard Summary
+          '34117-2' => false, # History and physical note
+          '28570-0' => false, # Procedure note
+          '11506-3' => false # Progress not
+        }
+
+        documentreference_ary&.each do |resource|
+          codings = resource&.type&.coding
+          next if codings.nil?
```
Jammjammjamm: When possible, perform some sort of filter operation rather than using `next` inside of `each`. In this case I'd probably do something like:
```
document_references
  .select { |document_reference| document_reference&.resource&.type&.coding&.present? }
  .each do |document_reference|
    ...
```
The nested eaches could do something similar.

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,228 @@
...
+      details %(
+
+        The #{title} Sequence tests DiagnosticReport and DocumentReference resources associated with the provided patient.  The resources
+        returned will be checked for consistency against the [US Core Clinical Notes Guidance](https://build.fhir.org/ig/HL7/US-Core-R4/clinical-notes-guidance.html)
+
+      )
+
+      @clinicalnotes_found = false
+
+      def test_clinicalnotes_documentreference(type_code)
```
Jammjammjamm: `clinical_notes`, `document_reference`

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: removed these two local variables

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: fixed

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: fixed

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: fixed

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: fixed

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: fixed

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: fixed

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,211 @@
...
+          '34117-2', # History and physical note
+          '28570-0', # Procedure note
+          '11506-3' # Progress not
+        ]
+
+        @actual_type_codes = Set[]
+
+        document_references&.select { |docref| docref&.type&.coding&.present? }&.each do |docref|
+          docref.type.coding.select { |coding| coding&.system == 'http://loinc.org' && @required_type_codes.include?(coding&.code) }&.each do |coding|
+            @actual_type_codes.add(coding.code)
```
Jammjammjamm: It's fine to use `add`, but it's more typical to use the "shovel operator" (it's really just a method rather than an operator): `@actual_type_codes << coding.code`

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,211 @@
...
+        resource_count = reply&.resource&.entry&.length || 0
+        @resources_found = true if resource_count.positive?
+
+        assert @resources_found, "Clinical Notes shall have at least one DiagnosticReport with category #{category_code}"
+
+        diagnostic_reports = reply&.resource&.entry&.map { |entry| entry&.resource }
+
+        diagnostic_reports&.each do |diarpt|
+          diarpt&.presentedForm&.select { |attachment| attachment&.url&.present? }&.each do |attachment|
+            assert @attachment_urls.key?(attachment.url), "Attachment #{attachment.url} referenced in DiagnosticReport/#{diarpt.id} but not in any DocumentReference"
```
Jammjammjamm: I see two issues with `@attachment_urls`:
- the name `flag` doesn't say what is being flagged
- it's called `@attachment_urls` but it's really just the attachment urls from docrefs

I'd address both of these by replacing `@attachment_urls` with two instance variables, `@docref_attachments` and `@report_attachments`. These variables could be hashes where the keys are attachment urls, and the values are the resource id `@docref_attachments[attachment.url] = docref.id`. Then this assertion could be removed from this test, and test 10 could do this for both sets of urls. Subtracting arrays gives their difference, so you can easily get a list of missing attachments.
```
irb(main):006:0> @docref_urls = ['url1', 'url2', 'url3', 'url4']
=> ["url1", "url2", "url3", "url4"]
irb(main):007:0> @report_urls = ['url3', 'url1']
=> ["url3", "url1"]
irb(main):008:0> @docref_urls - @report_urls
=> ["url2", "url4"]
```

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,211 @@
...
+      conformance_supports :DocumentReference, :DiagnosticReport
+
+      details %(
+
+        The #{title} Sequence tests DiagnosticReport and DocumentReference resources associated with the provided patient.  The resources
+        returned will be checked for consistency against the [US Core Clinical Notes Guidance](https://build.fhir.org/ig/HL7/US-Core-R4/clinical-notes-guidance.html)
+
+      )
+
+      @clinicalnotes_found = false
```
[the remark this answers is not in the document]
yunwwang: fixed

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: fixed

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,211 @@
...
+        assert_bundle_response(reply)
+
+        resource_count = reply&.resource&.entry&.length || 0
+        @resources_found = true if resource_count.positive?
+
+        assert @resources_found, "Clinical Notes shall have at least one DiagnosticReport with category #{category_code}"
+
+        diagnostic_reports = reply&.resource&.entry&.map { |entry| entry&.resource }
+
+        diagnostic_reports&.each do |diarpt|
```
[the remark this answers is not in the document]
yunwwang: fixed. Changed to diagnostic_report (singular)

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
yunwwang: Add Class ClinicalNoteAttachment to save resource_type and attachment hash.
Add function assert_attachment_matched to compare two attachment hashes and build error message

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,232 @@
...
+          '18842-5', # Dischard Summary
+          '34117-2', # History and physical note
+          '28570-0', # Procedure note
+          '11506-3' # Progress note
+        ]
+
+        @actual_type_codes = Set[]
+
+        document_references&.select { |document| document&.type&.coding&.present? }&.each do |document|
+          document.type.coding.select { |coding| coding&.system == 'http://loinc.org' && @required_type_codes.include?(coding&.code) }&.each do |coding|
```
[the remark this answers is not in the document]
yunwwang: US Core requires 5 types of document references must exists. There are more types than just these required 5. These looping section is to check if the current document reference has one of the required type. Document references that do not have required type is still valid. We just ignore it.

lib/app/modules/us_core_r4/clinicalnotes_sequence.rb
```diff
@@ -0,0 +1,232 @@
...
+
+        @required_type_codes = Set[
+          '11488-4', # Consultation Note
+          '18842-5', # Dischard Summary
+          '34117-2', # History and physical note
+          '28570-0', # Procedure note
+          '11506-3' # Progress note
+        ]
+
+        @actual_type_codes = Set[]
```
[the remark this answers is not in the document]
yunwwang: I give credit to Steve. He taught me that.

PR #250, opened 2019-07-24
Make authorization search params the same as first search

czh-orz:
FI-288

This PR sets the search parameters of the authorization tests to be the same as the first search's parameters. If there are no searches, then we won't do the authorization test.

### DIFF
--- a/generators/uscore-r4/generator.rb
+++ b/generators/uscore-r4/generator.rb
@@ -80,7 +80,7 @@ def create_authorization_test(sequence)
   }
 
   first_search = sequence[:searches].select { |search_param| search_param[:expectation] == 'SHALL' }.first ||
-                 sequence[:searches].select { |search_param| search_param[:expectation] == 'SHOULD' }.first
+                 sequence[:searches].find { |search_param| search_param[:expectation] == 'SHOULD' }
   return if first_search.nil?
 
   authorization_test[:test_code] = %(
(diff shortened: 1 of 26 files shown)


generators/uscore-r4/generator.rb
```diff
@@ -79,11 +79,15 @@ def create_authorization_test(sequence)
     link: 'http://www.fhir.org/guides/argonaut/r2/Conformance-server.html'
   }
 
+  first_search = sequence[:searches].select { |search_param| search_param[:expectation] == 'SHALL' }.first ||
```
Jammjammjamm:
```suggestion
  first_search = sequence[:searches].find { |search_param| search_param[:expectation] == 'SHALL' } ||
```

generators/uscore-r4/generator.rb
```diff
@@ -79,11 +79,15 @@ def create_authorization_test(sequence)
...
+                 sequence[:searches].select { |search_param| search_param[:expectation] == 'SHOULD' }.first
```
Jammjammjamm:
```suggestion
                 sequence[:searches].find { |search_param| search_param[:expectation] == 'SHOULD' }
```

lib/app/modules/us_core_r4/us_core_medication_sequence.rb
```diff
@@ -23,26 +23,9 @@ class UsCoreR4MedicationSequence < SequenceBase
 
       @resources_found = false
 
-      test 'Server rejects Medication search without authorization' do
```
Jammjammjamm: What happened to this test?

PR #253, opened 2019-07-25
Fi 253 add omit status

hershilpatel:
Add an omit result for a test in a sequence. An omitted test indicates that the test was omitted in the final sequence result (and is indicated by a grey circle)

To test this feature, set disable_tls_tests to true in config.yml, and run the app as normal.  The test 'FHIR server secured by transport layer security' in capability_statement_sequence.rb

Other updates:

The function skip_if_tls_disabled has been changed to omit_if_tls_disabled in skip_helpers.rb 

Updated sequence_result to handle omitted tests and make the rollup work correctly 

Added test cases for sequence_result

Will add sequence tests to test sequences as a whole will be added in a future feature 

image


test/model/sequence_result_test.rb
```diff
@@ -3,11 +3,310 @@
...
+    assert_equal 1, sequence_result.error_count
+    assert_equal 1, sequence_result.todo_count
+    assert_equal 1, sequence_result.skip_count
+    assert_equal 1, sequence_result.optional_passed
+    assert_equal 4, sequence_result.optional_total
+    assert_equal 1, sequence_result.required_omitted
+    assert_equal 1, sequence_result.optional_omitted
+  end
+
+  def test_sequence_result_result_pass
```
yunwwang: Can you synchronize the function names. For example, this one is test_sequence_result_result_pass. Another one is test_sequence_result_fail.

test/model/sequence_result_test.rb
```diff
@@ -3,11 +3,310 @@
...
+    test_result1.required = true
+    test_result1.result = Inferno::ResultStatuses::SKIP
+    sequence_result.test_results << test_result1
+
+    sequence_result.update_result_counts
+
+    assert_equal Inferno::ResultStatuses::SKIP, sequence_result.result
+  end
+
+  def test_sequence_result_result_optional_skip_when_no_pass_tests
```
yunwwang: In this test, sequence_result does not result "optional_skip". Should change to a better name?

test/model/sequence_result_test.rb
```diff
@@ -3,11 +3,310 @@
...
+    test_result2.required = false
+    test_result2.result = Inferno::ResultStatuses::ERROR
+    sequence_result.test_results << test_result2
+
+    sequence_result.update_result_counts
+
+    assert_equal Inferno::ResultStatuses::PASS, sequence_result.result
+  end
+
+  def test_sequence_result_pass_fail_vs_error
```
yunwwang: This function name is really confusing. What should be the result? Is it pass, or fail or error?

test/model/sequence_result_test.rb
hershilpatel: renamed to test_sequence_result_pass (and renamed all functions with "result_result")

test/model/sequence_result_test.rb
hershilpatel: changed name to test_sequence_result_pass_when_only_1_optional_skip_test

test/model/sequence_result_test.rb
hershilpatel: renamed test to test_sequence_result_error_when_pass_fail_and_error_tests_in_sequence

lib/app/views/test_case.erb
```diff
@@ -70,10 +70,15 @@
...
                     <%= test_case.sequence.test_count %> tests
                   <% else %>
-                    <%= sequence_results[test_case.id].required_passed %>/<%= sequence_results[test_case.id].required_total%> Required Tests Passed
+                    <%= sequence_results[test_case.id].required_passed %>/<%= sequence_results[test_case.id].required_total  - sequence_results[test_case.id].required_omitted %> Required Tests Passed
+
                     <% if sequence_results[test_case.id].optional_total > 0%> -
                       <%= sequence_results[test_case.id].optional_passed%>/<%= sequence_results[test_case.id].optional_total%> Optional Tests Passed
                     <% end%>
+
+                    <% if sequence_results[test_case.id].total_omitted > 0%> -
```
radamson:
```suggestion
                    <% if sequence_results[test_case.id].total_omitted.positive?%> -
```

test/model/sequence_result_test.rb
```diff
@@ -3,11 +3,310 @@
...
+    result.required_omitted = 4
+    result.optional_omitted = 300
+
+    assert_equal 304, result.total_omitted
+  end
+
+  def test_update_result_counts
+    sequence_result = Inferno::Models::SequenceResult.new
+
+    test_result1 = Inferno::Models::TestResult.new
```
Jammjammjamm: Could you DRY up these tests? Perhaps something like:

```
result_params = [
  { required: true, result: Inferno::ResultStatuses::PASS },
  { required: false, result: Inferno::ResultStatuses::PASS },
   ...
]
sequence_result.test_results = result_params.map { |params| Inferno::Models::TestResult.new(params) }
```

lib/app/views/test_case.erb
```diff
@@ -70,10 +70,15 @@
...
+                      <%= sequence_results[test_case.id].total_omitted %> Test <% if sequence_results[test_case.id].total_omitted != 1%>s<% end%> Omitted
```
radamson: If we are going to make Test singular or plural based on the number of tests admitted then we should be consistent and do it for required and omitted tests too.

lib/app/views/test_case.erb
```diff
@@ -70,10 +70,15 @@
                   <% if sequence_results[test_case.id].nil? %>
                     <%= test_case.sequence.test_count %> tests
                   <% else %>
-                    <%= sequence_results[test_case.id].required_passed %>/<%= sequence_results[test_case.id].required_total%> Required Tests Passed
+                    <%= sequence_results[test_case.id].required_passed %>/<%= sequence_results[test_case.id].required_total  - sequence_results[test_case.id].required_omitted %> Required Tests Passed
+
                     <% if sequence_results[test_case.id].optional_total > 0%> -
```
radamson:
```
<% if sequence_results[test_case.id].optional_total > 0%> -
  <%= sequence_results[test_case.id].optional_passed%>/<%= sequence_results[test_case.id].optional_total%> Optional Tests Passed
<% end%>
```

We repeat this a few times to handle required, optional and omitted tests.  Might be worth looping over it once for each of the different cases instead.

lib/app/models/sequence_result.rb
```diff
@@ -9,7 +10,7 @@ class SequenceResult
       include DataMapper::Resource
       property :id, String, key: true, default: proc { SecureRandom.uuid }
       property :name, String
-      property :result, String
+      property :result, String, default: ResultStatuses::PASS
```
radamson: I could see defaulting to pass causing some some issues to silently slip through.

hershilpatel:
FI-253

Posting Screenshot of feature

lib/app/models/sequence_result.rb
```diff
@@ -84,9 +100,9 @@ def update_result_counts
             end
           when ResultStatuses::SKIP
             if result.required
-              self.skip_count += 1
               self.result = result.result if pass?
             end
+            self.skip_count += 1
```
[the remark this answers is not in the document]
hershilpatel: I assume the skip_count should be counting all skips, not just the required ones.  If not we should rename it.  I did a "find in files" for all references for skip_count, and this change does not really affect anything because we only really use it for display.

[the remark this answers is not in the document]
radamson: I remember discussing this previously and I think we landed on only counting required skips even though the total wouldn't add up. 
 See FI-157 and PR-157.  @arscan might want to weigh in to make sure there won't be unintended consequences.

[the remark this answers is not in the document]
hershilpatel: ok for now I can change it back, but should we consider renaming it in case we later want to track optional_skips or total_skips?

test/model/sequence_result_test.rb
hershilpatel: Updated tests to follow this format

lib/tasks/tasks.rake
```diff
@@ -114,14 +114,18 @@ def execute(instance, sequences)
...
+      elsif result.error?
         print 'X error'.magenta
         print " - #{result.test_id} #{result.name}\n"
         puts "    Message: #{result.message}"
         print_requests(result).map do |req|
           puts "      #{req}"
         end
         fails = true
+      elsif result.omit?
+        print '* omit'.light_black
```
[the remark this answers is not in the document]
hershilpatel: yeah, I have never seen that before haha

lib/app/views/test_case.erb
hershilpatel: image

radamson: I was thinking of if `0/1 Optional Tests Passed` should be singular, but looking at it the way it is now is fine because of the fraction I think it's fine the way it is.

lib/app/models/sequence_result.rb
hershilpatel: I talked to Rob about this, what we were doing before this was explicitly setting the result to PASS on before calling update_result_counts (see line 124 in sequence_base), which which was the only thing causing update_result_counts to work at all (because otherwise it was returning nil). Long story short, we decided to set the default value to PASS for now, and there is a ticket in JIRA to make a PENDING status work

lib/app/views/test_case.erb
Jammjammjamm: For future reference, since we use `ActiveSupport`, we have access to the super handy `String#pluralize` method:
```
[1] pry(#<TestInstanceTest>)> 'test'.pluralize
=> "tests"
[2] pry(#<TestInstanceTest>)> 'test'.pluralize(0)
=> "tests"
[3] pry(#<TestInstanceTest>)> 'test'.pluralize(1)
=> "test"
[4] pry(#<TestInstanceTest>)> 'test'.pluralize(2)
=> "tests"
```

lib/app/models/sequence_result.rb
hershilpatel: For now, removed default value and instead the update_result_counts will set the result to PASS if it is nil after looping through all the test cases

radamson: This looks ok for now, but we need to address how we set results to pass everywhere more comprehensively in a future PR.  This will be especially important if we add more statuses like `pending.

hershilpatel:
fixed the count in the report and updated the icon legend

PR #254, opened 2019-07-25
Save R4 resources from the conformance statement

czh-orz:
FI-266
Added R4 resources so that they are saved from reading the conformance statement. 

To test this, run the MedicationRequest sequence in HSPC. It should now run the read, vread, and history tests.

### DIFF
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -163,6 +163,7 @@ module Inferno
         resources = ['Patient',
                      'AllergyIntolerance',
                      'CarePlan',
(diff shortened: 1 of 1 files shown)


radamson:
Looks good.

## MedicationRequest Test
**Before:**
Screen Shot 2019-07-30 at 11 14 14 AM
**After:**
Screen Shot 2019-07-30 at 11 14 40 AM

## Client State
**Before:**
Screen Shot 2019-07-30 at 11 15 23 AM
**After:**
Screen Shot 2019-07-30 at 11 15 09 AM

PR #255, opened 2019-07-25
added a rubocop config file for windows to allow windows users to run…

hershilpatel:
Added rubocop config file that windows users can run locally to avoid rubocop complaints of carriage returns.  This file merely inherits from the .rubocop.yml so it does not need to be updated separately

### DIFF
--- a/.rubocop-windows.yml
+++ b/.rubocop-windows.yml
@@ -1,0 +1,4 @@
+inherit_from: .rubocop.yml
+
+Layout/EndOfLine:
+  EnforcedStyle: crlf


hershilpatel:
@czh-orz to use this , use the following command (assuming your files are on windows but you are using wsl to run rubocop)

 bundle exec rubocop -c .rubocop-windows.yml

czh-orz:
It worked great for me, except lib/app/utils/logging.rb still triggered the carriage return thing for some reason. That was the only one though.

hershilpatel:
hmm that is strange. is there a branch I can checkout to look into this?

czh-orz:
I was just in this branch. I wouldn't worry about it though.

PR #258, opened 2019-07-26
FI-273: Refactor OAuth2 endpoints 

Jammjammjamm:
This branch adds happy path tests and separates `launch` and `redirect` into separate endpoints. When these endpoints successfully find a testing instance, they `pass` to a shared route which continues test execution.

### DIFF
--- a/lib/app/endpoint/oauth2_endpoints.rb
+++ b/lib/app/endpoint/oauth2_endpoints.rb
@@ -45,7 +45,7 @@ module Inferno
           end
 
           get '/oauth2/:key/launch/?' do
-            @instance = Inferno::Models::SequenceResult.recent_results_for_iss(params[:iss]).testing_instance
+            @instance = Inferno::Models::SequenceResult.recent_results_for_iss(params[:iss])&.testing_instance
             pass if @instance.present?
 
             @instance = Inferno::Models::TestingInstance.get(instance_id_from_cookie)
--- a/lib/app/endpoint/oauth2_endpoints.rb
+++ b/lib/app/endpoint/oauth2_endpoints.rb
@@ -8,67 +8,69 @@ module Inferno
           # Resume oauth2 flow
           # This must be early so it doesn't get picked up by the other routes
           get '/oauth2/:key/:endpoint/?' do
-            instance = nil
-            error_message = nil
-
             if params[:endpoint] == 'redirect'
-              instance = Inferno::Models::TestingInstance.first(state: params[:state])
-              if instance.nil?
-                instance = Inferno::Models::TestingInstance.get(cookies[:instance_id_test_set]&.split('/')&.first)
-                if instance.nil?
-                  error_message = %(
-                               <p>
-                                  Inferno has detected an issue with the SMART launch.
-                                  No actively running launch sequences found with a state of #{params[:state]}.
-                                  The authorization server is not returning the correct state variable and
-                                  therefore Inferno cannot identify which server is currently under test.
-                                  Please click your browser's "Back" button to return to Inferno,
-                                  and click "Refresh" to ensure that the most recent test results are visible.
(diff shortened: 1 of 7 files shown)


lib/app/endpoint/oauth2_endpoints.rb
```diff
@@ -5,91 +5,74 @@ class App
...
-              when 'stu3'
-                client.use_stu3
-              when 'dstu2'
-                client.use_dstu2
-              else
-                client.use_r4
-              end
-              client.default_json
-              sequence = test_case.sequence.new(instance, client, settings.disable_tls_tests, sequence_result)
+              client = FHIR::Client.for_testing_instance(@instance)
```
radamson: Would it be better to have something like a `get_client` method on SequenceBase  or TestingInstance instead of patching FHIR::Client?

lib/app/endpoint/oauth2_endpoints.rb
```diff
@@ -5,91 +5,74 @@ class App
...
-                  halt 500, message
-                elsif instance&.waiting_on_sequence&.wait?
-                  error_message = 'No iss for redirect'
-                else
-                  redirect "#{base_path}/#{cookies[:instance_id_test_set]}/?error=no_ehr_launch&iss=#{params[:iss]}"
-                end
-              end
+          get '/oauth2/:key/redirect/?' do
+            @instance = Inferno::Models::TestingInstance.first(state: params[:state])
+            pass if @instance.present?
```
radamson: This `pass` should be moved to the bottom of the route implementation and happen regardless of if the instance is present.  Having it here prevents us from rescuing the test with the cookie on the next line if a state parameter isn't passed (which is the cookies only job so far).

We should pass regardless, because the next route calls `running_test_found` and handles the checking whether `@instance.present` and provides the appropriate 500 response

lib/app/endpoint/oauth2_endpoints.rb
```diff
@@ -5,91 +5,74 @@ class App
...
+
+            @instance = Inferno::Models::TestingInstance.get(instance_id_from_cookie)
```
radamson: This should only happen if `instance.nil?` as it did previously.  Otherwise we'll overwrite the instance we correctly got from the state param.

lib/app/endpoint/oauth2_endpoints.rb
```diff
@@ -5,91 +5,74 @@ class App
...
+            elsif @instance&.waiting_on_sequence&.wait?
+              @error_message = "State provided in redirect (#{params[:state]}) does not match expected state (#{@instance.state})."
+            else
+              redirect "#{base_path}/#{cookies[:instance_id_test_set]}/?error=no_state&state=#{params[:state]}"
+            end
+          end
+
+          get '/oauth2/:key/launch/?' do
+            @instance = Inferno::Models::SequenceResult.recent_results_for_iss(params[:iss])&.testing_instance
+            pass if @instance.present?
```
radamson: This `pass` should be at the bottom too.

lib/app/endpoint/oauth2_endpoints.rb
```diff
@@ -5,91 +5,74 @@ class App
...
+
+            @instance = Inferno::Models::TestingInstance.get(instance_id_from_cookie)
```
radamson: Should be moved into the `if @instance.nil?` block below.

lib/app/endpoint/oauth2_endpoints.rb
Jammjammjamm: I added more tests to cover these cases and adjusted the code accordingly, although not exactly how you suggested.

radamson: Awesome, I like the way you handled it here.

PR #265, opened 2019-07-29
Fix capitalization of sequence titles

czh-orz:
FI-267

Fix capitalization of sequences by looking at the title in the profile definition.


generators/uscore-r4/metadata_extractor.rb
```diff
@@ -34,6 +33,8 @@ def extract_metadata
 
   def build_new_sequence(resource, profile)
     base_name = profile.split('StructureDefinition/')[1]
+    profile_json = get_json_from_uri(profile_uri(base_name))
+    profile_title = profile_json['title'].gsub('US ', '').gsub('Core ', '').gsub('Profile', '').strip
```
yunwwang:
```suggestion
    profile_title = profile_json['title'].gsub(/US\s*Core\s*/, '').gsub(/\s*Profile/, '').strip
```
Us regular expression

czh-orz: Fixed!

czh-orz:
> Could you update the class names to start with `USCore` rather than `UsCore`?

Should be fixed now

PR #267, opened 2019-07-31
Fix FI-297 Add ClinicalNotes sequence to us_ocre_module.yml automatic…

yunwwang:
Add ClinicalNotesSequence to module.yml.rb template.
Fix class name from UsCoreR4ClincialNotesSequence to USCoreR4ClinicalNotesSequence

### DIFF
--- a/generators/uscore-r4/templates/module.yml.erb
+++ b/generators/uscore-r4/templates/module.yml.erb
@@ -23,3 +23,4 @@ test_sets:
         sequences:<% sequences.each do |sequence| %>
         - <%=sequence[:classname]%><% end %>
         - R4ProvenanceSequence
+        - USCoreR4ClinicalNotesSequence
--- a/lib/app/modules/us_core_module.yml
+++ b/lib/app/modules/us_core_module.yml
@@ -46,3 +46,4 @@ test_sets:
         - USCoreR4PractitionerroleSequence
         - USCoreR4ProcedureSequence
         - R4ProvenanceSequence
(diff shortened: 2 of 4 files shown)


radamson:
Looks good to me.  Nice catch on capitalizing `US`.

@czh-orz Are there instructions anywhere on how to run the test generator?  Just wanted to put this through its paces.

generators/uscore-r4/templates/module.yml.erb
```diff
@@ -23,3 +23,4 @@ test_sets:
         sequences:<% sequences.each do |sequence| %>
         - <%=sequence[:classname]%><% end %>
         - R4ProvenanceSequence
```
radamson: As an aside for a future PR we should consider if Provenance tests belong in the US Core module.  I didn't see any Provenance requirements in the [IG](https://www.hl7.org/fhir/us/core/index.html)

Jammjammjamm: Provenance isn't there yet, but it will be: https://confluence.hl7.org/display/SEC/Basic+Provenance+Implementation+Guide

czh-orz:
No, the instructions aren't up anywhere. You just need to run "ruby generator.rb" in the generator directory.

generators/uscore-r4/templates/module.yml.erb
yunwwang: Here is the latest CI version of US Core R4:
http://build.fhir.org/ig/argonautproject/R4/branches/master/basic-provenance.html

PR #268, opened 2019-08-06
FI-289 removed tooltip on warning icon indicating redirect

hershilpatel:
Removed tooltip because it was conflicting for another one (as in issue 289)

### DIFF
--- a/lib/app/views/test_list.erb
+++ b/lib/app/views/test_list.erb
@@ -39,7 +39,7 @@
       <% end %>
       <% if result.test_warnings.length > 0 %>
           <div class="result-details-icon result-details-icon-warning" data-toggle="tooltip" title="<%= result.test_warnings.length %> warning(s).  Warnings do not result in a test failure.">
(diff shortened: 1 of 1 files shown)


lib/app/views/test_list.erb
```diff
@@ -39,7 +39,7 @@
       <% end %>
       <% if result.test_warnings.length > 0 %>
           <div class="result-details-icon result-details-icon-warning" data-toggle="tooltip" title="<%= result.test_warnings.length %> warning(s).  Warnings do not result in a test failure.">
-            <span class="oi oi-warning" data-toggle="tooltip" title="Test is waiting for a server launch or redirect"></span>
```
Jammjammjamm: This fixes the double tooltip, but was this other tooltip still needed in other cases?

hershilpatel: In every scenario both tooltips will show up (because they both show up if result.test_warnings.length > 0) so if we do need this tool tip, it needs to be moved. My guess is that now we have multiple types of warning so we show the warning icon with the more general tooltip, and then they can click results to actually see the warning messages

PR #270, opened 2019-08-08
Fi 292 order nonpatient search sequences

czh-orz:
This will save references to non-patient search sequences found in other searches. Non-patient search sequences will have a prerequisite for an id for the resource.


generators/uscore-r4/generator.rb
```diff
@@ -31,8 +31,22 @@ def generate_search_validators(metadata)
 end
 
 def generate_tests(metadata)
+  # first isolate the profiles that don't have patient searches
+  delayed_sequences = metadata[:sequences].select do |sequence|
```
Jammjammjamm: This seems like it could be a separate method.

generators/uscore-r4/templates/module.yml.erb
```diff
@@ -18,9 +18,13 @@ test_sets:
         - ManualRegistrationSequence
         - StandaloneLaunchSequence
         - EHRLaunchSequence
-      - name: US Core R4
+      - name: US Core R4 Patient Based Profiles
         run_all: true
-        sequences:<% sequences.each do |sequence| %>
+        sequences:<% sequences.reject { |seq| seq[:delayed_sequence] }.each do |sequence| %>
```
Jammjammjamm: Could this be assigned to a variable so that the logic doesn't live in the template?

generators/uscore-r4/templates/module.yml.erb
```diff
@@ -18,9 +18,13 @@ test_sets:
...
         - <%=sequence[:classname]%><% end %>
         - R4ProvenanceSequence
         - USCoreR4ClinicalNotesSequence
+      - name: US Core R4 Non-Patient Profiles
+        run_all: true
+        sequences:<% sequences.select { |seq| seq[:delayed_sequence] }.each do |sequence| %>
```
Jammjammjamm: Same here.

lib/app/models/testing_instance.rb
```diff
@@ -52,6 +52,12 @@ class TestingInstance
 
       property :dynamic_registration_token, String
 
+      property :practitioner, String
+      property :practitionerrole, String
+      property :medication, String
+      property :organization, String
+      property :location, String
```
Jammjammjamm: Could there be a more general solution here? Manually maintaining a list of special resources in `TestingInstance` and the database schema seems problematic.

lib/app/sequence_base.rb
```diff
@@ -712,6 +712,24 @@ def validate_reference_resolutions(resource)
         assert(problems.empty?, problems.join("<br/>\n"))
       end
 
+      def save_delayed_sequence_references(resource)
+        non_patient_sequences = ['Location', 'Organization', 'Medication', 'Practitioner', 'PractitionerRole']
```
Jammjammjamm: We really should avoid globally defining resources for special treatment. This is especially problematic since this list exists here and in `TestingInstance`. Rather than maintaining this list, I think each non-patient sequence should declare itself to be one.

generators/uscore-r4/generator.rb
czh-orz: Moved to another function.

generators/uscore-r4/templates/module.yml.erb
czh-orz: Moved to variable

lib/app/sequence_base.rb
czh-orz: made it so sequences mark themselves as delayed

lib/app/models/testing_instance.rb
czh-orz: Made it so the properties are added in the sequences.

generators/uscore-r4/generator.rb
```diff
@@ -62,6 +70,19 @@ def generate_tests(metadata)
   end
 end
 
+def mark_delayed_sequences(metadata)
```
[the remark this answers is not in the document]
czh-orz: Yes, this is how the sequences know to mark themselves.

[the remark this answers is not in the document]
Jammjammjamm: Ok, so this is in the generator, and this is what tells a sequence to include `mark_delayed`?

lib/app/modules/us_core_r4/us_core_organization_sequence.rb
```diff
@@ -11,7 +11,8 @@ class USCoreR4OrganizationSequence < SequenceBase
 
       test_id_prefix 'Organization' # change me
 
-      requires :token, :patient_id
+      mark_delayed :Organization
```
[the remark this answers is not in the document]
czh-orz: I made it so it just gets the resource from the conformance_supports declaration.

generators/uscore-r4/generator.rb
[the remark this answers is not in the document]
czh-orz: That's right

lib/app/modules/us_core_r4/us_core_organization_sequence.rb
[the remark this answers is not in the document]
czh-orz: Made it so it finds the resources during the save_delayed_sequence_references step so that it's not dependent on being after conformance_supports.

Jammjammjamm:
I think this will be good if tests are added for `save_delayed_sequence_references`.

lib/app/sequence_base.rb
```diff
@@ -712,6 +729,24 @@ def validate_reference_resolutions(resource)
         assert(problems.empty?, problems.join("<br/>\n"))
       end
 
+      def save_references_for_delayed_sequences(resource)
+        delayed_resource_types = @@conformance_supports.select { |sequence, _resources| @@delayed_sequences.include? sequence }.values.flatten
+        walk_resource(resource) do |value, meta, _path|
+          next if meta['type'] != 'Reference'
+
+          begin
```
Jammjammjamm: Extraneous `begin`?

test/unit/sequence_base_test.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require_relative '../test_helper'
+
+class DelayedResourceTestSequence < MiniTest::Test
```
arscan: Why is this named Delayed Resource Test Sequence if it is in `test/unit/sequence_base_test.rb`?

test/unit/sequence_base_test.rb
```diff
@@ -0,0 +1,51 @@
...
+      scopes: 'launch openid patient/*.* profile',
+      token: 99_897_979
+    )
+
+    @instance.save!
+
+    client = FHIR::Client.new(@instance.url)
+    client.use_r4
+    client.default_json
+    @sequence = Inferno::Sequence::USCoreR4AllergyintoleranceSequence.new(@instance, client, true)
```
arscan: I know it is convenient to reference IG-specific sequences in our tests, but at some point we need to split this apart so that the platform doesn't depend on the existence of sequences.  Not a problem to fix now but it would be good for everyone to start thinking about Inferno the platform vs. IG-specific tests.

test/unit/sequence_base_test.rb
czh-orz: I started writing it as a unit test for just the delayed stuff, but forgot to change the name when I renamed the file. Should be fixed now.

test/unit/sequence_base_test.rb
czh-orz: That's true. Luckily it still worked when I just changed the AllergyIntolerance sequence to SequenceBase.

test/sequence/us_core_r4/us_core_r4_allergy_intolerance_sequence_test.rb
```diff
@@ -0,0 +1,129 @@
...
+    )
+
+    @instance.save!
+
+    @instance.resource_references << Inferno::Models::ResourceReference.new(
+      resource_type: 'Patient',
+      resource_id: @patient_id
+    )
+
+    @instance.supported_resources << Inferno::Models::SupportedResource.create(
```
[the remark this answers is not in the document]
arscan: Is this resolved?

lib/app/sequence_base.rb
arscan: Is this resolved?

PR #271, opened 2019-08-12
FI-290: Dynamically determine supported resources

Jammjammjamm:
Currently, Inferno maintains a global list of the resources tested across all of the current modules. This is a maintenance issue because adding sequences for other resources requires changing this list in `TestingInstance` in addition to adding the new sequence. Currently, the report shows the number of supported resources that the server supports based on this global list (currently 25 resources), not based on the number of resources actually required for the current module (14 for the resource in the screenshot below):
Screen Shot 2019-08-12 at 11 01 13 AM

Meanwhile, the list of resource interactions in the "State" modal displays a list of 20 resources:
Screen Shot 2019-08-12 at 11 03 35 AM

This PR removes the global resource support list and instead determines the necessary resource types for each module based on the `conformance_supports` statements in each sequence. The report now only displays the number of resources required for the current module.
Before fetching the capability statement:
Screen Shot 2019-08-12 at 11 08 34 AM
After fetching the capability statement:
Screen Shot 2019-08-12 at 11 09 22 AM

The "State" modal also now only displays the resources required for the module:
Screen Shot 2019-08-12 at 11 10 33 AM

Behind the scenes, the `SupportedResource` model has been removed and replaced with the `ServerCapabilities` model. This model keeps a copy of the server's capability statement and dynamically determines the capabilities from it rather than maintaining persistent state about each supported resource.

This PR also moves the `module` class into `lib/app/models` and breaks out its contained classes into separate files.


lib/app/models/testing_instance.rb
```diff
@@ -2,6 +2,7 @@
```
czh-orz: interaction_supported? is not used anymore

lib/app/models/module.rb
```diff
@@ -2,139 +2,31 @@
...
+    attr_accessor :resources
     attr_accessor :test_sets
+    attr_accessor :title
 
-    def initialize(name, description, default_test_set, fhir_version, title)
-      @name = name
-      @description = description
-      @default_test_set = default_test_set
-      @fhir_version = fhir_version
+    def initialize(**params)
```
hershilpatel: Why  switch out explicit parameters for an map of parameters?

Jammjammjamm: In the existing code, the call site was performing logic to combine two parameters into one, and also contained some rather complex logic to perform additional initialization of the module. I wanted to move all that logic into the constructor, which would have meant increasing the parameter count from five to seven. Five parameters is already pretty bad, but seven is really bad. The parameters were already in a hash at the call site, so it seemed undesirable to pull them out and put them in a particular order rather than just passing the hash.

hershilpatel: Makes sense!

PR #272, opened 2019-08-16
FI-291: Fix Sort Order of Extended Sequences

okeefm:
### Problem: 
`TestResult`s used to display in a non-obvious/nondeterministic order (see the "before" screenshot below) after a test was cancelled. This is because of how test indexes are assigned to base tests in extended sequences. See ticket https://oncprojectracking.healthit.gov/support/browse/FI-291 for more information.

### Solution:  
No longer reassign test indexes for the base tests when they're added to extended sequences, e.g. the `CapabilityStatementSequence` tests in the `ArgonautConformanceStatement` sequence. See the "after" screenshots below for examples of the Cancelled EHR Launch Sequence test results, a completed test suite, and a test suite that hasn't yet been run, to show that they are all sorted as expected.

### Before:
Screen Shot 2019-08-16 at 10 45 00 AM

### After:
Cancelled test suite:
Screen Shot 2019-08-16 at 8 21 10 AM

Completed test suite:
Screen Shot 2019-08-16 at 8 24 28 AM

Not-run test suite:
Screen Shot 2019-08-16 at 8 24 12 AM

### DIFF
--- a/lib/app/modules/argonaut/argonaut_conformance_sequence.rb
+++ b/lib/app/modules/argonaut/argonaut_conformance_sequence.rb
@@ -128,7 +128,7 @@ module Inferno
 
       test 'Conformance Statement lists supported Argonaut profiles, operations and search parameters' do
         metadata do
-          id '05'
+          id '06'
           link 'https://www.fhir.org/guides/argonaut/r2/Conformance-server.html'
           desc %(
            The Argonaut Data Query Implementation Guide states:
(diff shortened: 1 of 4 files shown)


Jammjammjamm:
Could we have the database do this sort rather than ruby?

okeefm:
I just gave that a try, and it does seem to do the job (changing the `TestResult` association in `SequenceResult` to `has n, :test_results, order: [:test_id.asc]`). I'll switch the PR around to do that instead.

okeefm:
@arscan Rebased into `development`. After some discussion in the lab, the fix for this now removes a line from `extends_sequence` in `sequence_base`, which is something you added in a refactor. Are you able to re-review to ensure that we haven't broken something in fixing this ordering issue?

PR #277, opened 2019-08-23
FI-304: Create production docker images

Jammjammjamm:
This PR modifies the Inferno Dockerfile so that the docker image contains the Inferno code in addition to its dependencies.

A dockerhub repository has been set up so that docker images will be built on any push/merge to development: https://cloud.docker.com/u/onchealthit/repository/docker/onchealthit/inferno-dev

I created a fork to test the auto-build process: https://cloud.docker.com/u/smacvicar/repository/docker/smacvicar/inferno-test
You can test this prebuilt image by running:
```
docker pull smacvicar/inferno-test
docker run -p 4567:4567 smacvicar/inferno-test
```


.ruby-version
```diff
@@ -0,0 +1 @@
+2.5.5
```
radamson: Why ruby 2.5.5 specifically instead of 2.6.4 (or 2.5.6 even).  If we have to use a specific version we should at least use the latest patch I imagine.

I'm fine with locking a version, but just testing it out I had `rbenv` throw me `rbenv: version '2.5.5' is not installed`

radamson: Another thought here is that if we push towards making Inferno into a gem we'll want it to support multiple versions and this won't make as much sense.  Treating it as an application this does make it more reproducible though.

.travis.yml
```diff
@@ -1,9 +1,6 @@
 language: ruby
 services:
   - docker
-rvm:
-  - 2.5
-  - 2.6
```
radamson: Should we remove the CI builds for different versions of Ruby?  The builds for multiple versions let us see if anything broke between versions.

Multiple Build Example PR for reference: https://travis-ci.org/onc-healthit/inferno/builds/580890307

lib/app/endpoint.rb
```diff
@@ -20,6 +20,7 @@ class Endpoint < Sinatra::Base
       Inferno::EXTRAS = settings.include_extras
 
       if settings.logging_enabled
+        $stdout.sync = true # output in Docker is heavily delayed without this
```
radamson: Should this be independently configurable?  I'm not sure if there is a performance impact with doing this, but we might not always care about stdout anyways.  What do you think?

Dockerfile
```diff
@@ -1,8 +1,28 @@
-FROM ruby:2.5
+FROM ruby:2.5.5
```
radamson: If we are specifying a patch version should it be the latest (2.5.6) to include all the patches?

.ruby-version
Jammjammjamm: That is the version that is currently being used in production. If you're using a different version locally, you'll have to install the specified version.

Jammjammjamm: I think that's a bridge we should cross when we come to it. Right now Inferno is an application, and I think we should ensure that the production and development environments use the same ruby version.

lib/app/endpoint.rb
Jammjammjamm: If we're expecting to run Inferno in docker in production (which I think we are), we will want this. I don't think it's worth adding configuration so that we could avoid a potential performance penalty only in development.

.travis.yml
Jammjammjamm: This slows down the tests and I don't think we are currently gaining any value from it. I think the version being used in production is the only one that really matters.

radamson: We've found some differences before with it, but if we're going to only use a single version of Ruby then I agree this is an unnecessary step.

okeefm: It might be worth adding in when we decide to upgrade to Ruby 2.6, but it's probably ok out for now.

Jammjammjamm: Travis respects your `.ruby-version` file, so we still won't need to specify a ruby version here.

Dockerfile
```diff
@@ -1,8 +1,28 @@
...
+RUN bundle install
+
+### Install Inferno
+
+RUN mkdir data
+COPY public /var/www/inferno/public
+COPY resources /var/www/inferno/resources
+COPY config* /var/www/inferno/
+COPY Rakefile /var/www/inferno/
+COPY test /var/www/inferno/test
```
okeefm: Is this directory copied over so we can do testing on the Docker image?

Jammjammjamm: Yes. I'd really prefer to keep all the tests and testing dependencies out of it, but no good solutions for that came to mind. Something to keep in mind for the future.

PR #281, opened 2019-08-27
FI-311 bulk data patient

yunwwang:
Code changes for initial test sequences including metadata and patient/$export

### DIFF
--- a/lib/app/models/server_capabilities.rb
+++ b/lib/app/models/server_capabilities.rb
@@ -28,6 +28,10 @@ module Inferno
         end
       end
 
+      def operation_supported?(op_name)
+        statement.rest.any? { |r| r.operation.any? { |x| x.name == op_name } }
+      end
+
       private
 
       def statement
--- a/lib/app/models/server_capabilities.rb
+++ b/lib/app/models/server_capabilities.rb
@@ -28,8 +28,8 @@ module Inferno
         end
       end
 
-      def operation_supported?(op_name)
-        statement.rest.any? { |r| r.operation.any? { |x| x.name == op_name } }
+      def operation_supported?(operation_name)
+        statement.rest.any? { |rest| rest.operation.any? { |operation| operation.name == operation_name } }
       end
 
       private
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -187,11 +187,6 @@ module Inferno
         end
       end
 
-      def operation_supported?(op_name)
-        operation = server_capabilities.capabilities.rest&.first&.operation&.select { |x| x.name == op_name }
-        operation.present?
-      end
-
       def save_resource_reference(type, id, profile = nil)
         resource_references
           .select { |ref| (ref.resource_type == type) && (ref.resource_id == id) }
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -187,6 +187,11 @@ module Inferno
         end
       end
 
+      def operation_supported?(op_name)
+        operation = server_capabilities.capabilities.rest&.first&.operation&.select { |x| x.name == op_name }
+        operation.present?
+      end
+
       def save_resource_reference(type, id, profile = nil)
         resource_references
           .select { |ref| (ref.resource_type == type) && (ref.resource_id == id) }
--- a/lib/app/modules/bulk_data/bulk_data_capability_statement_sequence.rb
+++ b/lib/app/modules/bulk_data/bulk_data_capability_statement_sequence.rb
@@ -90,6 +90,22 @@ module Inferno
         formats = ['json', 'applcation/json', 'application/json+fhir', 'application/fhir+json']
         assert formats.any? { |format| @conformance.format.include? format }, 'Conformance does not state support for json.'
       end
+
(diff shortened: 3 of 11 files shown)


lib/app/modules/bulk_data/bulk_data_capability_statement_sequence.rb
```diff
@@ -0,0 +1,95 @@
...
+        * SMART on FHIR endpoints
+
+        For more information of the Capability Statement, visit these links:
+
+        * [Capability](https://www.hl7.org/fhir/capabilitystatement.html)
+        * [Argonaut Capability Requirements](http://hl7.org/fhir/us/core/2019Jan/CapabilityStatement-us-core-server.html)
+        * [SMART on FHIR Conformance](http://hl7.org/fhir/smart-app-launch/conformance/index.html)
+      )
+
+      test 'FHIR server capability states JSON support' do
```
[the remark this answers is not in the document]
Jammjammjamm: What about checking `$export` support?

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -0,0 +1,75 @@
...
+        end
+
+        reply = export_kick_off('Patient')
+
+        # Shall return 202
+        assert_response_accepted(reply)
+
+        @content_location = reply.response[:headers]['content-location']
+
+        # Shall have Content-location
```
Jammjammjamm: I don't think this comment is necessary.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -0,0 +1,75 @@
...
+      title 'Patient Tests'
+
+      description 'Verify that Patient resources on the Bulk Data server follow the Bulk Data Access Implementation Guide'
+
+      test_id_prefix 'Patient'
+
+      requires :token
+      conformance_supports :Patient
+
+      # export
```
Jammjammjamm: I don't think this comment is necessary.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -0,0 +1,75 @@
...
+          id '02'
+          link 'https://build.fhir.org/ig/HL7/bulk-data/export/index.html#bulk-data-kick-off-request'
+          desc %(
+          )
+          versions :stu3
+        end
+
+        reply = export_kick_off('Patient')
+
+        # Shall return 202
```
Jammjammjamm: If this comment is necessary to explain what the assertion is asserting, then `assert_response_code` should be used instead so that you can be explicit about what status you're accepting and won't need a comment to explain the assertion.

test/sequence/bulk_data/patient_export_sequence_test.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    @content_location = 'http://www.example.com/status'
+
+    client = FHIR::Client.new(@instance.url)
+    client.use_stu3
+    client.default_json
+    @sequence = Inferno::Sequence::BulkDataPatientExportSequence.new(@instance, client, true)
+  end
+
+  def include_export_sub(code = 202, headers = { content_location: @content_location })
```
Jammjammjamm: `include_export_stub`

lib/app/models/testing_instance.rb
```diff
@@ -187,6 +187,11 @@ def conformance_supported?(resource, methods = [])
         end
       end
 
+      def operation_supported?(op_name)
+        operation = server_capabilities.capabilities.rest&.first&.operation&.select { |x| x.name == op_name }
```
Jammjammjamm: This is only testing the server-level operations. I think you also need to test the resource-level operations (at `rest.resource.operations`) since the IG supports `export` for Patient and Group. I feel like this should use `any?` rather than `first` and `select`?

Also, could you use a descriptive variable name rather than `x`?

lib/app/models/server_capabilities.rb
```diff
@@ -28,6 +28,10 @@ def supported_interactions
         end
       end
 
+      def operation_supported?(op_name)
+        statement.rest.any? { |r| r.operation.any? { |x| x.name == op_name } }
```
Jammjammjamm: Please use descriptive variable names rather than `r` and `x`.

lib/app/models/testing_instance.rb
yunwwang: This method is redundant. Removed from the latest fix.

Jammjammjamm: The version in `ServerCapabilities` has the same problem. It's only checking server-level operations, but (correct me if I'm wrong) `Patient` and `Group` exports are resource-level operations, so it needs to check operations at the resource level (i.e. check `capabilities.rest.resource.operation` rather than only `capabilities.rest.operation`).

yunwwang: According to the IG capability statement (http://hl7.org/fhir/uv/bulkdata/STU1/CapabilityStatement-bulk-data.json.html), those operations are declared in rest ```rest``` level. Your comment also make sense. I would like to keep align with current IG and I will contact IG author to figure out if IG need to be updated.

PR #284, opened 2019-09-03
create rake task for test generator

czh-orz:
To test, run "bundle exec rake generator:us_core_r4" in terminal.

### DIFF
--- a/lib/tasks/tasks.rake
+++ b/lib/tasks/tasks.rake
@@ -788,3 +788,11 @@ namespace :terminology do |_argv|
     Inferno::Terminology.create_validators(validator_type)
   end
 end
+
+namespace :generator do |_argv|
+  desc 'Generate US Core R4 Tests'
+  task :us_core_r4 do
+    path = File.expand_path('../../generators/uscore-r4/generator.rb', __dir__)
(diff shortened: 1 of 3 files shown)


lib/tasks/tasks.rake
```diff
@@ -788,3 +788,11 @@ namespace :terminology do |_argv|
     Inferno::Terminology.create_validators(validator_type)
   end
 end
+
+namespace :generator do |_argv|
+  desc 'Generate US Core R4 Tests'
+  task :us_core_r4 do
+    path = File.expand_path('../../generators/uscore-r4/generator.rb', __dir__)
+    system('ruby', path)
```
Jammjammjamm: This is where it would be really nice if the generator were a class so that you could say something like `Generator.run` rather than having to use `system`.

arscan:
I think a better API may be 

```bundle exec rake inferno:generate[us_core_r4]```

But I'm good with merging this in because at least it moves un in the right direction, and we can create two new tickets to address the naming and turning it into a class instead of a standalone script.

Jammjammjamm:
I'll make those tickets.

PR #285, opened 2019-09-03
FI-320 make SHOULD searches optional

czh-orz:
Searches that are marked as "SHOULD" will be optional.

### DIFF
--- a/generators/uscore-r4/generator.rb
+++ b/generators/uscore-r4/generator.rb
@@ -98,7 +98,8 @@ def create_search_test(sequence, search_param)
   search_test = {
     tests_that: "Server returns expected results from #{sequence[:resource]} search by #{search_param[:names].join('+')}",
     index: sequence[:tests].length + 1,
-    link: 'https://build.fhir.org/ig/HL7/US-Core-R4/CapabilityStatement-us-core-server.html'
+    link: 'https://build.fhir.org/ig/HL7/US-Core-R4/CapabilityStatement-us-core-server.html',
+    optional: search_param[:expectation] == 'SHOULD'
   }
 
   is_first_search = search_test[:index] == 2 # if first search - fix this check later
--- a/generators/uscore-r4/generator.rb
+++ b/generators/uscore-r4/generator.rb
@@ -99,7 +99,7 @@ def create_search_test(sequence, search_param)
     tests_that: "Server returns expected results from #{sequence[:resource]} search by #{search_param[:names].join('+')}",
     index: sequence[:tests].length + 1,
     link: 'https://build.fhir.org/ig/HL7/US-Core-R4/CapabilityStatement-us-core-server.html',
-    optional: search_param[:expectation] == 'SHOULD'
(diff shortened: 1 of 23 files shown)


generators/uscore-r4/generator.rb
```diff
@@ -98,7 +98,8 @@ def create_search_test(sequence, search_param)
   search_test = {
     tests_that: "Server returns expected results from #{sequence[:resource]} search by #{search_param[:names].join('+')}",
     index: sequence[:tests].length + 1,
-    link: 'https://build.fhir.org/ig/HL7/US-Core-R4/CapabilityStatement-us-core-server.html'
+    link: 'https://build.fhir.org/ig/HL7/US-Core-R4/CapabilityStatement-us-core-server.html',
+    optional: search_param[:expectation] == 'SHOULD'
```
Jammjammjamm: Are `SHALL` and `SHOULD` the only options? Would it be more flexible if it were `!= SHALL` instead?

generators/uscore-r4/templates/sequence.rb.erb
```diff
@@ -26,7 +26,8 @@ module Inferno
       test '<%=test[:tests_that]%>' do
         metadata do
           id '<%=format('%02d', test[:index])%>'
-          link '<%=test[:link]%>'
+          link '<%=test[:link]%>'<%='
+          optional' if test[:optional]%>
```
Jammjammjamm: Having the erb tag split over multiple lines looks really weird to me. I'd definitely prefer to have the whole tag on one line, or to do `<% if test[:optional] %>optional<% end %>`.

lib/app/modules/us_core_r4/us_core_practitionerrole_sequence.rb
```diff
@@ -49,8 +49,9 @@ def validate_resource_item(resource, property, value)
         @client.set_no_auth
         skip 'Could not verify this functionality when bearer token is not set' if @instance.token.blank?
 
-        specialty_val = @practitionerrole&.specialty&.coding&.first&.code
+        specialty_val = resolve_element_from_path(@practitionerrole, 'specialty.coding.code')
```
Jammjammjamm: Where are these changes coming from? I don't see them reflected in the generator diff.

generators/uscore-r4/generator.rb
czh-orz: You're right. Made it to `!= SHALL`

generators/uscore-r4/templates/sequence.rb.erb
czh-orz: Couldn't make it one line due to newline issues. Used the ` if test[:optional]` approach.

lib/app/modules/us_core_r4/us_core_practitionerrole_sequence.rb
czh-orz: That line should've been changed in a previous PR, but for some reason it didn't git committed. Running the generator again just added the changes back in.

PR #287, opened 2019-09-05
FI-302: Fix broken SMART assertion

Jammjammjamm:
This PR fixes an assertion that couldn't fail and DRYs up the conformance tests a little.

### DIFF
--- a/lib/app/modules/argonaut/argonaut_conformance_sequence.rb
+++ b/lib/app/modules/argonaut/argonaut_conformance_sequence.rb
@@ -118,12 +118,9 @@ module Inferno
 
         assert_valid_conformance
 
-        extensions = @conformance.try(:rest).try(:first).try(:security).try(:extension)
-        assert !extensions.nil?, 'No SMART capabilities listed in conformance.'
-        capabilities = extensions.select { |x| x.url == 'http://fhir-registry.smarthealthit.org/StructureDefinition/capabilities' }
-        assert !capabilities.nil?, 'No SMART capabilities listed in conformance.'
-        available_capabilities = capabilities.map(&:valueCode)
-        missing_capabilities = (required_capabilities - available_capabilities)
+        assert @server_capabilities.smart_support?, 'No SMART capabilities listed in conformance.'
+
+        missing_capabilities = (required_capabilities - @server_capabilities.smart_capabilities)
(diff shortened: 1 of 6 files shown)


lib/app/modules/argonaut/argonaut_conformance_sequence.rb
```diff
@@ -115,14 +116,11 @@ class ArgonautConformanceSequence < CapabilityStatementSequence
...
+
+        assert @server_capabilities.smart_support?, 'No SMART capabilities listed in conformance.'
 
-        extensions = @conformance.try(:rest).try(:first).try(:security).try(:extension)
-        assert !extensions.nil?, 'No SMART capabilities listed in conformance.'
-        capabilities = extensions.select { |x| x.url == 'http://fhir-registry.smarthealthit.org/StructureDefinition/capabilities' }
-        assert !capabilities.nil?, 'No SMART capabilities listed in conformance.'
-        available_capabilities = capabilities.map(&:valueCode)
-        missing_capabilities = (required_capabilities - available_capabilities)
+        missing_capabilities = (required_capabilities - @server_capabilities.smart_capabilities)
```
Jammjammjamm: I think this might be outdated. Based on my reading of the SMART spec, [the CapabilityStatement is used to communicate the OAuth endpoints, but not the individual SMART capabilities](http://hl7.org/fhir/smart-app-launch/conformance/index.html#smart-on-fhir-core-capabilities-and-capability-sets). If others agree that this is the case, I will remove this code and add a ticket to create a test which checks the well-known endpoint for these capabilities.

arscan: Yes, this appears to have been removed.  Which makes sense, nobody used this yet, and SMART only kept the auth URIs in there for backwards compatibility for existing apps.  Looks like it was commented out and I couldn't find any references to the extension anywhere in the most recently published SMART spec (though the extension does exist in the [FHIR spec](https://www.hl7.org/fhir/extension-capabilities.html) still, that should probably be removed).

Capture

Do you want to merge this in and then we fix this in another PR, or should we hold off on merging and fix this problem here?

Jammjammjamm: I'd like to fix it in this PR.

PR #289, opened 2019-09-05
Fi 312 bulk data status check

yunwwang:
Add test step for export status check request


lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -14,14 +14,27 @@ class BulkDataPatientExportSequence < SequenceBase
...
+        @content_location = reply.response[:headers]['content-location']
 
-        @client.get(url, @client.fhir_headers(headers))
+        assert @content_location.present?, 'Export response header did not include "Content-Location"'
+      end
+
+      def assert_export_status(content_location = @content_location)
+        reply = export_status_check(content_location)
+
+        assert reply.code == 200, "Bad response code: expected 200, 202, but found #{reply.code}."
```
arscan: Your description does not match the value checked.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -59,13 +72,55 @@ def export_kick_off(klass)
           versions :stu3
```
arscan: Bulk data requires `r4` because async doesn't in exist prior versions.

lib/app/utils/assertions.rb
```diff
@@ -101,9 +101,9 @@ def assert_resource_content_type(client_reply, content_type)
       response_content_type = header
       response_content_type = header[0, header.index(';')] unless header.index(';').nil?
 
-      return if assertion_negated(response_content_type == "application/fhir+#{content_type}")
```
arscan: If we are fixing this, also rename the assertion to `assert_response_content_type_equals`.  Having `resource` in the method name makes no sense, and since it is checking the content_type is something specific that is passed in, we need to be more specific. We have to be careful about changing the behavior of shared assertions, but it doesn't appear that this one is used anywhere else.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -59,13 +72,55 @@ def export_kick_off(klass)
...
+        url += '/$export'
+
+        @client.get(url, @client.fhir_headers(headers))
+      end
+
+      def export_status_check(url, headers = { accept: 'application/json' })
+        wait_time = 1
+        reply = nil
+
+        loop do
```
arscan: Could we add some kind of guard against waiting indefinitely?  It could be hardcoded to something reasonable now (skip if wait time > 10 minutes or something), but eventually it should be an input that is configured and passed to teh sequence like other parameters.  In practice we will have to put bounds on reasonable response times in our test scenario, even if there isn't an actual requirement stating things have to be done within a certain amount of time.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -59,13 +72,55 @@ def export_kick_off(klass)
...
+          reply = @client.get(url, @client.fhir_headers(headers))
+
+          break if reply.code != 202
+
+          retry_after = reply.response[:headers]['retry-after']
+          wait_time = (retry_after.presence || wait_time * 2).to_i
```
arscan: What if they have no retry-after header?  This will cause `retry_after` to be nil, and `retry_after.presence` to raise an unhelpful "no presence method on nil on line xyz" error (vs. a nicely worded failure).  Similarly what if presence isn't a number.  Or if presense in in milliseconds but we expect seconds... will a huge number here cause this test to hang indefinitely?

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
yunwwang: The status check response shall be either 202 (still processing) or 200 (ready). Status 202 is handled inside export_status_check function. The logic inside that function is that if the status code is 202 then wait and check status again. If the status code is not 202 then return. This assert function here only need to check 200.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
yunwwang: Both Smart-health reference server and BCDA test server are STU3.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
yunwwang: retry_after.presence return nil even if the 'retry-after' header does NOT exist.
Add a logic check if (current time used + wait time > timeout)

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
arscan: But `prefer: respond-async` isn't standardized until `R4`.   I guess you could argue that they are STU3 servers doing extra stuff.  Regardless of whether it is technically valid in STU3, this definitely should be applicable to R4, R5, R6+++, so just remove that line altogether so that it isn't tied to one version.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
arscan: Ah, ok.  Its good that it prints out something that would be logical to the user, but on the surface it looks like an oversight in the code.  I think a comment is necessary here would save confusion later.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -59,13 +74,68 @@ def export_kick_off(klass)
...
+        headers = { accept: 'application/json' }
+        start = Time.now
 
-        assert @content_location.present?, 'Server response must have "Content-Location" in header for $export request'
+        loop do
+          reply = @client.get(url, @client.fhir_headers(headers))
+
+          wait_time = get_wait_time(wait_time, reply)
+          seconds_used = Time.now - start + wait_time
+
```
arscan: Regarding the timeouts --would it be easier to at this point do

```skip "Server took more than #{timeout} seconds to process the request."  if seconds_used > timeout``` 

And then get rid of the other timeout skipping code earlier?  It is hard for me to follow the logic between timeouts and skipping over there, because it is not obvious why you would skip when `status == 202`.  I trust that it works, it just that it will take me some time to figure why it would work.

yunwwang: export_status_check function just check status. It does not handle any assert/skip logic. The logic here is that if the status is still 202 and pass the timeout threshold, export_status_check stops next status check and returns status 202 to caller. The caller, assert_export_status handles assert/logic based on the returned status code. If the code is 200, the status check completed. If the code is 202, the status check timed out. Otherwise, there is server error. 
The function export_status_check may finally merged into fhirclient module as common bulk data export function

PR #290, opened 2019-09-05
Validate against correct profiles

czh-orz:
This will validate resources against the correct profile. To test, you can run the Smoking Status sequence. Before, it was saying the code shouldn't be 72166-2.

### DIFF
--- a/generators/uscore-r4/generator.rb
+++ b/generators/uscore-r4/generator.rb
@@ -10,6 +10,19 @@ require_relative './metadata_extractor'
 OUT_PATH = '../../lib/app/modules'
 RESOURCE_PATH = '../../resources/us_core_r4/'
 
+PROFILE_URIS = {
+  smoking_status: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-smokingstatus.json',
+  diagnostic_report_lab: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-diagnosticreport-lab.json',
+  diagnostic_report_note: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-diagnosticreport-note.json',
+  observation_lab: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-observation-lab.json',
+  pediatric_bmi_age: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-pediatric-bmi-for-age.json',
+  pediatric_weight_height: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-pediatric-weight-for-height.json'
+}.freeze
+
+def validation_profile_uri(sequence)
+  "Inferno::ValidationUtil::US_CORE_R4_URIS[:#{PROFILE_URIS.key(sequence[:profile])}]" if PROFILE_URIS.key(sequence[:profile])
+end
+
 def run
   redownload_files = (ARGV&.first == '-d')
   FileUtils.rm Dir.glob("#{RESOURCE_PATH}*") if redownload_files
@@ -116,7 +129,7 @@ def create_search_test(sequence, search_param)
 
         @#{sequence[:resource].downcase} = reply.try(:resource).try(:entry).try(:first).try(:resource)
         @#{sequence[:resource].downcase}_ary = reply&.resource&.entry&.map { |entry| entry&.resource }
-        save_resource_ids_in_bundle(versioned_resource_class('#{sequence[:resource]}'), reply)
+        save_resource_ids_in_bundle(versioned_resource_class('#{sequence[:resource]}'), reply#{', ' + validation_profile_uri(sequence) if validation_profile_uri(sequence)})
         validate_search_reply(versioned_resource_class('#{sequence[:resource]}'), reply, search_params))
     else
       %(
@@ -216,7 +229,7 @@ def create_resource_profile_test(sequence)
   }
   test[:test_code] = %(
         skip 'No resources appear to be available for this patient. Please use patients with more information.' unless @resources_found
-        test_resources_against_profile('#{sequence[:resource]}'))
+        test_resources_against_profile('#{sequence[:resource]}'#{', ' + validation_profile_uri(sequence) if validation_profile_uri(sequence)}))
 
   sequence[:tests] << test
 end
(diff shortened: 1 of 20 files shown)


lib/app/utils/validation.rb
```diff
@@ -54,6 +54,15 @@ def self.get_resource(json, version)
       snf: 'https://bluebutton.cms.gov/assets/ig/StructureDefinition-bluebutton-snf-claim'
     }.freeze
 
+    US_CORE_R4_URIS = {
```
Jammjammjamm: I think this file is very problematic because it tightly couples core Inferno behavior to specific sequences. I don't know that it needs to be fixed in this PR, but I think a better solution would be for `SequenceBase` to have a default implementation that just looks at the profile the resource declares, and for sequences to override that method as needed (probably by creating and including a module).

lib/app/utils/validation.rb
```diff
@@ -133,7 +142,23 @@ def self.guess_r4_profile(resource)
       return if resource.blank?
 
       candidates = RESOURCES[:r4][resource.resourceType]
-      return candidates.first if candidates.present?
+      return if candidates.blank?
```
Jammjammjamm: This method needs to be unit tested.

generators/uscore-r4/generator.rb
```diff
@@ -10,6 +10,19 @@
 OUT_PATH = '../../lib/app/modules'
 RESOURCE_PATH = '../../resources/us_core_r4/'
 
+PROFILE_URIS = {
+  smoking_status: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-smokingstatus.json',
+  diagnostic_report_lab: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-diagnosticreport-lab.json',
+  diagnostic_report_note: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-diagnosticreport-note.json',
+  observation_lab: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-observation-lab.json',
```
Jammjammjamm: I think this should be `lab_results`.

generators/uscore-r4/generator.rb
```diff
@@ -10,6 +10,19 @@
 OUT_PATH = '../../lib/app/modules'
 RESOURCE_PATH = '../../resources/us_core_r4/'
 
+PROFILE_URIS = {
+  smoking_status: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-smokingstatus.json',
```
Jammjammjamm: Why are different urls used here and in `ValidationUtil`?

generators/uscore-r4/generator.rb
```diff
@@ -10,6 +10,19 @@
...
+  diagnostic_report_lab: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-diagnosticreport-lab.json',
+  diagnostic_report_note: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-diagnosticreport-note.json',
+  observation_lab: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-observation-lab.json',
+  pediatric_bmi_age: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-pediatric-bmi-for-age.json',
+  pediatric_weight_height: 'https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-pediatric-weight-for-height.json'
+}.freeze
+
+def validation_profile_uri(sequence)
+  "Inferno::ValidationUtil::US_CORE_R4_URIS[:#{PROFILE_URIS.key(sequence[:profile])}]" if PROFILE_URIS.key(sequence[:profile])
```
Jammjammjamm: I think it's worth assigning `PROFILE_URIS.key(sequence[:profile])` to variable.

lib/app/utils/validation.rb
```diff
@@ -133,7 +142,23 @@ def self.guess_r4_profile(resource)
...
+
+      if resource.resourceType == 'Observation'
+        return DEFINITIONS[US_CORE_R4_URIS[:smoking_status]] if resource&.code&.coding&.any? { |coding| coding&.code == '72166-2' }
+
+        return DEFINITIONS[US_CORE_R4_URIS[:observation_results]] if resource&.category&.coding&.any? { |coding| coding&.code == 'laboratory' }
```
Jammjammjamm: This key isn't in `US_CORE_R4_URIS`.

lib/app/utils/validation.rb
```diff
@@ -133,7 +142,23 @@ def self.guess_r4_profile(resource)
...
+
+        return DEFINITIONS[US_CORE_R4_URIS[:pediatric_bmi_age]] if resource&.code&.coding&.any? { |coding| coding&.code == '59576-9' }
+
+        return DEFINITIONS[US_CORE_R4_URIS[:pediatric_weight_height]] if resource&.code&.coding&.any? { |coding| coding&.code == '77606-2' }
+      elsif resource.resourceType == 'DiagnosticReport'
+        return DEFINITIONS[US_CORE_R4_URIS[:diagnostic_report_lab]] if resource&.category&.coding&.any? { |coding| coding&.code == 'LAB' }
+
+        return DEFINITIONS[US_CORE_R4_URIS[:diagnostic_report_note]]
```
Jammjammjamm: Doesn't this still need a condition?

generators/uscore-r4/generator.rb
```diff
@@ -116,7 +129,7 @@ def create_search_test(sequence, search_param)
 
         @#{sequence[:resource].downcase} = reply.try(:resource).try(:entry).try(:first).try(:resource)
         @#{sequence[:resource].downcase}_ary = reply&.resource&.entry&.map { |entry| entry&.resource }
-        save_resource_ids_in_bundle(versioned_resource_class('#{sequence[:resource]}'), reply)
+        save_resource_ids_in_bundle(versioned_resource_class('#{sequence[:resource]}'), reply#{', ' + validation_profile_uri(sequence) if validation_profile_uri(sequence)})
```
Jammjammjamm: I think this would be nicer if this logic happened outside of the string. Something like:
```
save_resource_ids_in_bundle_arguments = [
  "versioned_resource_class('#{sequence[:resource]}')",
  'reply',
  validation_profile_uri(sequence)
].compact.join(', ')
```

generators/uscore-r4/generator.rb
czh-orz: These are the profile URI's from the build version of US Core which the generator uses. The ValidationUtil checks against the published version.

Jammjammjamm: But why are different versions used?

lib/app/utils/validation.rb
czh-orz: So I'm still not sure about this. There isn't an easy restriction for diagnostic report note like there is for diagnostic report lab. So in the end, I just decided that if every diagnostic report is either a lab or a note for our testing purposes, then this approach would work. If the diagnostic report isn't a note or lab, then our tests won't be able to handle that anyway. But we should definitely discuss if there's a better way to handle this.

generators/uscore-r4/generator.rb
czh-orz: I don't think the published version existed when I started the generator. It looks like it was made on June 19.

lib/app/utils/validation.rb
czh-orz: Added a validation test

generators/uscore-r4/generator.rb
czh-orz: changed to lab_results

generators/uscore-r4/generator.rb
czh-orz: fixed

lib/app/utils/validation.rb
czh-orz: fixed

generators/uscore-r4/generator.rb
czh-orz: fixed

generators/uscore-r4/generator.rb
Jammjammjamm: But they exist now, so what purpose do these urls serve?

lib/app/utils/validation.rb
Jammjammjamm: I'll make a ticket for this.

lib/app/utils/validation.rb
Jammjammjamm: Yeah, there doesn't seem to be a different way to differentiate them.

PR #294, opened 2019-09-17
Fi 339 kick off header tests

yunwwang:

**Submitter:**
This pull request added two tests for invalid headers in export request
FI-339
https://oncprojectracking.healthit.gov/support/browse/FI-339

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -15,14 +15,24 @@ class BulkDataPatientExportSequence < SequenceBase
...
+        reply = export_kick_off(klass: klass)
 
         assert_response_accepted(reply)
         @content_location = reply.response[:headers]['content-location']
 
         assert @content_location.present?, 'Export response header did not include "Content-Location"'
       end
 
+      def assert_export_kick_off_fail_invalid_accept(klass)
+        reply = export_kick_off(klass: klass, headers: { accept: 'application/fhir+xml', prefer: 'respond-async' })
```
Jammjammjamm: I don't think these assertions should have side effects. I think the assertions should just be asserting something rather than carrying out the entire test. These tests are simple enough that all this code could just live in the test.

yunwwang: These are the test logic. I separate these logic from the test step, i.e. "test 'Server shall reject for $export operation with invalid Accept header'", so the logic can be easily unit tested. These are temporary solution before the test object approach

Jammjammjamm: It's fine to separate out that logic, but I don't think it belongs in an assertion. The assertion should just test the result of some previous steps, not perform the entire test.

yunwwang: I cannot using "test_" as prefix. Maybe I will rename to "check_"?

yunwwang: fixed

PR #295, opened 2019-09-17
FI-345: Update SMART tests

Jammjammjamm:
This PR updates the SMART tests to match the current spec. This main change is that SMART capabilities are now listed in the well-known endpoint rather than the server metadata.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-345

**Reviewer 1:**

Name: Rob Scanlon

**Reviewer 2:**

Name:


test/sequence/onc_program/onc_smart_discovery_sequence_test.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require_relative '../../test_helper'
+class ONCProgramProvenanceTest < MiniTest::Test
```
arscan:
```suggestion
class ONCSMARTDiscoveryTest < MiniTest::Test
```

lib/app/modules/smart/smart_discovery_sequence.rb
```diff
@@ -108,25 +156,28 @@ class SMARTDiscoverySequence < SequenceBase
...
+          id '05'
+          link 'http://hl7.org/fhir/smart-app-launch/conformance/index.html#using-cs'
           desc %(
-
-           If a server requires SMART on FHIR authorization for access, its metadata must support automated discovery of OAuth2 endpoints
-
+           If a server requires SMART on FHIR authorization for access, its
+           metadata must support automated discovery of OAuth2 endpoints
           )
         end
```
radamson: I know you didn't touch this part, but is there a reason we don't also check that the registration endpoint is consistent between the well-known endpoint response and the conformance statement? Seems like we should check that too.

lib/app/modules/smart/smart_discovery_sequence.rb
```diff
@@ -10,71 +10,119 @@ class SMARTDiscoverySequence < SequenceBase
...
+      end
 
-        @instance.update(oauth_authorize_endpoint: @well_known_authorize_url, oauth_token_endpoint: @well_known_token_url, oauth_register_endpoint: register_url)
+      test 'Configuration from well-known endpoint contains required fields' do
+        metadata do
+          id '02'
+          link 'http://hl7.org/fhir/smart-app-launch/conformance/index.html#metadata'
+          desc %(
+            The JSON from .well-known/smart-configuration contains the following
+            required fields: #{REQUIRED_WELL_KNOWN_FIELDS.join(', ')}
```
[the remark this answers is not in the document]
Jammjammjamm: I agree and will look into this. I had tried to do some emphasis here, but there's some weirdness around doing programmatic formatting in the descriptions.

lib/app/modules/smart/smart_discovery_sequence.rb
```diff
@@ -10,71 +10,119 @@ class SMARTDiscoverySequence < SequenceBase
...
+          optional
           desc %(
+            The JSON from .well-known/smart-configuration contains the following
+            recommended fields: #{RECOMMENDED_WELL_KNOWN_FIELDS.join(', ')}
+          )
+        end
 
-           If a server requires SMART on FHIR authorization for access, its metadata must support automated discovery of OAuth2 endpoints
+        missing_fields = RECOMMENDED_WELL_KNOWN_FIELDS - @well_known_configuration.keys
+        assert missing_fields.empty?, "The following recommended fields are missing: #{missing_fields.join(', ')}"
```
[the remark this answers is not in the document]
Jammjammjamm: Does it need to be wrapped in a warning when the entire test is optional?

lib/app/modules/smart/smart_discovery_sequence.rb
Jammjammjamm: I imagine that it's because this test is only testing the required endpoints.

lib/app/modules/smart/smart_discovery_sequence.rb
```diff
@@ -10,71 +10,119 @@ class SMARTDiscoverySequence < SequenceBase
...
+        @well_known_configuration = JSON.parse(well_known_configuration_response.body)
+        @well_known_authorize_url = @well_known_configuration['authorization_endpoint']
+        @well_known_token_url = @well_known_configuration['token_endpoint']
+        @instance.update(
+          oauth_authorize_endpoint: @well_known_authorize_url,
+          oauth_token_endpoint: @well_known_token_url,
+          oauth_register_endpoint: @well_known_configuration['registration_endpoint']
+        )
+
+        assert @well_known_configuration.present?, 'No .well-known/smart-configuration body'
```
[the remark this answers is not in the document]
Jammjammjamm: It had to do with guaranteeing that the instance variables were set and @instance was updated based on the contents of the well-known endpoint. Consider the following workflow:

- No body is received, so the assertion fails and the oauth urls aren't set
- Later, the urls are set by the conformance sequence
- This test is rerun, no body is received, the assertion fails, and the sequence continiues with the urls received from the conformance sequence
- The tests continue as though these urls were received from the well-known endpoint, leading to erroneous passes

I think I reorganized some code since I encountered this problem, and this code as is may not work as intended, so I will look into this some more.

lib/app/modules/smart/smart_discovery_sequence.rb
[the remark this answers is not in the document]
radamson: Good point, didn't think about it being optional.

lib/app/modules/smart/smart_discovery_sequence.rb
radamson: That makes sense, but I think it would be good to highlight it to folks so that they are aware in case those URLs are different even if we don't use them directly ourselves.  Doesn't have to be in this PR though.

lib/app/modules/smart/smart_discovery_sequence.rb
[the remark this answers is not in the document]
radamson: Closing this.  Figuring out how to make our descriptions look better seems like a job for another PR.

lib/app/modules/smart/smart_discovery_sequence.rb
```diff
@@ -10,129 +10,216 @@ class SMARTDiscoverySequence < SequenceBase
...
+
+        @well_known_configuration = JSON.parse(well_known_configuration_response.body)
+        @well_known_authorize_url = @well_known_configuration['authorization_endpoint']
+        @well_known_token_url = @well_known_configuration['token_endpoint']
+        @well_known_register_url = @well_known_configuration['registration_endpoint']
+        @well_known_manage_url = @well_known_configuration['management_endpoint']
+        @well_known_introspect_url = @well_known_configuration['introspection_endpoint']
+        @well_known_revoke_url = @well_known_configuration['revocation_endpoint']
+
+        @instance.update(
```
arscan: It seems that we should be saving the SMART .well-known file like we save the capability statement.  I was expecting to see some of this info show up in the state modal, but we are only saving a few of these fields.  Push this off to investigate in another ticket?

lib/app/modules/smart/smart_discovery_sequence.rb
```diff
@@ -10,129 +10,216 @@ class SMARTDiscoverySequence < SequenceBase
...
+      test 'Configuration from well-known endpoint contains recommended fields' do
         metadata do
-          id '02'
-          link 'http://www.hl7.org/fhir/smart-app-launch/capability-statement/'
+          id '03'
+          link 'http://hl7.org/fhir/smart-app-launch/conformance/index.html#metadata'
+          optional
           desc %(
+            The JSON from .well-known/smart-configuration contains the following
+            recommended fields: #{RECOMMENDED_WELL_KNOWN_FIELDS.map { |field| "`#{field}`" }.join(', ')}
```
arscan:
```suggestion
            recommended fields: #{RECOMMENDED_WELL_KNOWN_FIELDS.map { |field| "`#{field}`" }.join(', ')}.
            
            This test is optional because these fields are recommended, not required.
```

lib/app/modules/smart/smart_discovery_sequence.rb
```diff
@@ -10,129 +10,216 @@ class SMARTDiscoverySequence < SequenceBase
...
+        assert missing_fields.empty?, "The following recommended fields are missing: #{missing_fields.join(', ')}"
+      end
 
+      test 'Conformance/Capability Statement provides OAuth 2.0 endpoints' do
+        metadata do
+          id '04'
+          link 'http://hl7.org/fhir/smart-app-launch/conformance/index.html#using-cs'
+          desc %(
+           If a server requires SMART on FHIR authorization for access, its
+           metadata must support automated discovery of OAuth2 endpoints
```
arscan:
```suggestion
           metadata must support automated discovery of OAuth2 endpoints.
```

lib/app/modules/smart/smart_discovery_sequence.rb
```diff
@@ -10,129 +10,216 @@ class SMARTDiscoverySequence < SequenceBase
...
-          id '03'
-          link 'http://www.hl7.org/fhir/smart-app-launch/capability-statement/'
+          id '05'
+          link 'http://hl7.org/fhir/smart-app-launch/conformance/index.html#using-cs'
           desc %(
-
-           If a server requires SMART on FHIR authorization for access, its metadata must support automated discovery of OAuth2 endpoints
-
+           If a server requires SMART on FHIR authorization for access, its
+           metadata must support automated discovery of OAuth2 endpoints
```
arscan:
```suggestion
           metadata must support automated discovery of OAuth2 endpoints.
```

lib/app/modules/smart/smart_discovery_sequence.rb
Jammjammjamm: Seems like a good idea to me.

PR #297, opened 2019-09-20
Fi-323: Fix FHIR version checking

Jammjammjamm:
Currently, when trying to run a a module with no version specified (such as the SMART module), tests will be skipped because the FHIR versions don't match. This branch updates the version checking to properly support modules with no version requirement.

I added Minitest as an explicit dependency so that we aren't just getting it through one of our other dependencies. I also discovered that MiniTest supports spec-style tests, so I converted the TestingInstance tests. I think this style makes it a lot easier to write small tests with a clear intention and am looking for feedback from the team.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-323

**Reviewer 1:**

Name:

**Reviewer 2:**

Name: Rob Scanlon
- [X] Code is maintainable and reusable, reuses existing code and infrastructure
- [X] The tests appropriately test the new code, including edge cases
- [X] You have tried to break the code


test/unit/test_instance_test.rb
```diff
@@ -2,51 +2,76 @@
...
-    Inferno::Models::ServerCapabilities.create(
-      testing_instance_id: @instance.id,
-      capabilities: conformance.as_json
-    )
-
-    assert !@instance.conformance_supported?(:Patient, [:read])
-    assert !@instance.conformance_supported?(:Patient, [:read, :search, :history])
-    assert !@instance.conformance_supported?(:Patient)
-    assert @instance.conformance_supported?(:Observation, [:read])
+      assert !@instance.conformance_supported?(:Patient)
```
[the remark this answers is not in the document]
Jammjammjamm: I also don't have a strong opinion.

Jammjammjamm:
It would be nice to have one way to determine whether FHIR versions match, but I think that might be out of scope for this ticket. Maybe I should give the method a more specific name to make it clear that it's only for checking whether the instance and sequence match.

Gemfile
```diff
@@ -11,6 +11,7 @@ gem 'dm-sqlite-adapter'
 gem 'fhir_client'
 gem 'json-jwt'
 gem 'kramdown'
+gem 'minitest'
```
arscan: Should this be in the `group :test`?

arscan:
Also, this test approach does seem more explicit, though we might be able to accomplish the same thing using the old test format, but just writing them better :)

Gemfile
Jammjammjamm: I'm pretty sure this is where we get `assert`, so it's used outside of tests.

Jammjammjamm:
This may just be my ignorance, but with the other style tests, I don't know how to do scoped setup. I often find it useful to do things like:
```
describe SomeClass do
  before { global_setup}

  describe '#some_method' do
    before { setup_just_for_testing_this_method }
  ...
```

Gemfile
radamson: The `assert` in the sequences comes from the `Inferno::Assertions` module if that's what you're referring to.

https://github.com/onc-healthit/inferno/blob/dbb146a7bef896790c6e4695dde6520406740039/lib/app/utils/assertions.rb#L6-L10

Jammjammjamm: Well, that seems obvious in retrospect.

arscan:
> This may just be my ignorance, but with the other style tests, I don't know how to do scoped setup.

I suppose you would just call a setup method at the beginning of the test?  Yeah it isn't very clear from a semantic perspective, but you could address by how you name the method?  I am not very opinionated on this though.

PR #303, opened 2019-09-25
FI-377 Make test objects

Jammjammjamm:
This branch creates objects for each test in a Sequence as a first step towards making sequences more testable.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-330

**Reviewer 1:**

Name: Rob S

**Reviewer 2:**

Name:


lib/app/models/inferno_test.rb
```diff
@@ -0,0 +1,64 @@
...
+        @versions ||= versions
+      end
+
+      def metadata_hash
+        {
+          test_id: id,
+          name: name,
+          description: desc,
+          required: required?,
+          url: link
```
Jammjammjamm: Confusing!

lib/app/models/inferno_test.rb
```diff
@@ -0,0 +1,64 @@
+module Inferno
+  module Sequence
+    class InfernoTest
```
Jammjammjamm: Opinions on a better name?

lib/app/sequence_base.rb
```diff
@@ -367,119 +362,69 @@ def self.extends_sequence(klass)
...
-
-        current_test = @@test_metadata[sequence_name].last
-
-        wrapped = lambda do
-          instance_eval(&block) if @metadata_only # just run the test to hit the metadata block
+        tests << InfernoTest.new(name, @@test_index, &block)
+      end
 
+      def wrap_test(test)
+        lambda do |sequence|
```
Jammjammjamm: Any ideas where this lambda's argument is coming from?

lib/app/models/inferno_test.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      def optional?
+        @optional
+      end
+
+      def required?
+        !optional?
+      end
+
+      def desc(description = nil)
```
Jammjammjamm: This could be changed to `description` now to be consistent with Sequence descriptions, but it would require updating every test.

lib/tasks/tasks.rake
```diff
@@ -184,7 +185,14 @@ namespace :inferno do |_argv|
       csv << ['', '', '', '', '']
       csv << ['Test ID', 'Reference', 'Sequence/Group', 'Test Name', 'Required?', 'Reference URI']
       flat_tests.each do |test|
-        csv << [test[:test_id], test[:ref], test[:sequence].split('::').last, test[:name], test[:sequence_required] && test[:required], test[:url]]
+        csv << [
+          test[:test_id],
+          test[:ref],
```
Jammjammjamm: Is `ref` actually a thing?

Jammjammjamm: I stick a `pry` in here and it wasn't populated for any tests.

lib/app/models/inferno_test.rb
arscan: Force pushing completely confuses GitHub for these reviews ;-)

lib/app/models/inferno_test.rb
arscan: Does SequenceTest sound better?

lib/app/sequence_base.rb
arscan: What do you mean?  You don't use that argument anywhere inside the lambda, and you aren't passing it into the lambda when calling it.  Do you want to use that argument but can't figure out how to pass it in?  (I don't see why you don't just remove the argument)

lib/app/models/inferno_test.rb
arscan: I think I made it that way to keep the DSL a little less verbose within each test (maybe?) but I do kinda disagree with that approach now.  One thing we could do is allow either (and maybe mark it for deprecation?).  We should wait until a major version before we yank it regardless.

lib/tasks/tasks.rake
arscan: This is a placeholder for a reference back to regulation, if applicable.  Presumably this should be filled out with NPRM content for the Program Edition at this point, but we haven't done that.  I created a ticket for doing this when the rule is finalized.

lib/app/sequence_base.rb
Jammjammjamm: What you are saying is exactly how I understand it, but lambdas enforce the number of arguments and if I remove the argument, I get an `ArgumentError`. I could probably just replace the lambda with a Proc, but I was hoping to be able to remove the argument from wherever it's coming from instead.

arscan:
I mostly follow your new method for grabbing the metadata on the first run. Getting rid of the `metadata_only` argument is definitely a big improvement, so as long as it works I'm good with it 👍

lib/app/models/inferno_test.rb
Jammjammjamm: I think it seems about the same. Maybe it should just be `Test`? `Inferno::Sequence::Test`

arscan: My only objection to that is that Test is fairly generic and used by other libraries.  But that is weak.  Sequence::Test seems nice to me.  My real problem is that what we define as a 'Test' may defy expectations a bit because they behave differently based on context.   But that can't really be solved here.

lib/app/sequence_base.rb
arscan: See https://stackoverflow.com/questions/15304485/why-does-instance-eval-succeed-with-a-proc-but-not-with-a-lambda

`instance_eval` passes the instance of the object as the first parameter.  This seems to get in some gnarly aspects of binding blocks of code to objects.  Accepting `_sequence` seems fine by me.  Or you can change to `Proc` probably.  Or you can try `instance_exec`.  But there might be side effects that prevent those other options working the way we want (so I think its just best to leave as-is because it works).

Jammjammjamm: Thank you! `instance_exec` solved the problem, and I think it is the cleanest solution.

lib/app/models/inferno_test.rb
okeefm: I agree with @arscan here. Do we have a way to mark things for deprecation?

lib/app/sequence_base.rb
```diff
@@ -54,15 +54,14 @@ class SequenceBase
...
-      def initialize(instance, client, disable_tls_tests = false, sequence_result = nil, metadata_only = false)
+      def initialize(instance, client, disable_tls_tests = false, sequence_result = nil)
         @client = client
         @instance = instance
         @client.set_bearer_token(@instance.token) unless @client.nil? || @instance.nil? || @instance.token.nil?
         @client&.monitor_requests
         @sequence_result = sequence_result
         @disable_tls_tests = disable_tls_tests
         @test_warnings = []
-        @metadata_only = metadata_only
```
okeefm: Did we get rid of the idea of "metadata only" tests?

lib/app/models/inferno_test.rb
Jammjammjamm: I think we'd basically need to have some log output here stating that the method is deprecated.

lib/app/sequence_base.rb
Jammjammjamm: Yes, rather than an initial run of each test with `@metadata_only` set, each test now has an initial run in the context of an `InfernoTest` (instead of in the context of the Sequence the test belongs to. `@metadata_only` caused the `metadata` method to raise an exception after its block was evaluated so that the actual test didn't run. Now that is just how the `metadata` method in `InfernoTest` always works, while `metadata` in `SequenceBase` is a no-op.

arscan: Yeah this is neat.

lib/app/models/inferno_test.rb
arscan: fhir_models has a mixin for that:

https://github.com/fhir-crucible/fhir_models/blob/master/lib/fhir_models/deprecate.rb

Not sure if it is worthwhile to handle it generically like this or not, but it is a neat solution.  Note that this was contributed by Kevin from McKesson.

Jammjammjamm: Yeah, that's nice, but I probably wouldn't bring something like that in until we have multiple deprecations.

lib/app/views/test_details.erb
```diff
@@ -10,19 +10,12 @@
...
 
-    <% unless @test_metadata[:description].nil? %>
+    <% if @test.description.present? %>
       <div class="test-result-details-description">
-        <%= markdown_to_html(@test_metadata[:description]) %>
-        <a class="test-result-details-url"target="_blank" href="<%= @test_metadata[:url] %>"><%= @test_metadata[:url] %></a>
-      </div>
-    <% end %>
-
-    <% unless @test_metadata[:details].nil? %>
```
Jammjammjamm: I didn't see anywhere where this would be populated.

lib/app/models/inferno_test.rb
yunwwang: TestStep?

okeefm: what do MiniTest/Rspec call them?

arscan: I have been noodling with calling each of these 'steps' instead of 'tests'.  But that would be a pretty big change and perhaps this is something we can just keep in the back of our minds. Let's keep this as an 'InfernoTest' for now...

lib/app/views/test_details.erb
okeefm: Anywhere where which would be populated? `@test_metadata[:details]`?

arscan: Hmmm, there can be details on sequences (which there almost always are), and there can be details on result (which is an area that I would like to see improvement in... we just don't save enough from the assertions), but I'm not seeing anywhere that details could be on a test.  So I think this was a mistake or some kind of cruft left behind from a previous refactor.  I couldn't find anywhere to trigger this either.  So I'm good removing it.

lib/app/models/inferno_test.rb
arscan: `caller(0..0)` doesn't provide helpful information, seems like `caller(1..1)` is more useful because it shows the sequence where this is called.  If the `Test` class had a reference back to the sequence it is a part of we could avoid system calls, which make me a little nervous for use on other platforms we don't actively test.  But if that is messy then I'm good with this.

lib/app/models/test.rb
```diff
@@ -0,0 +1,79 @@
...
+      def required?
+        !optional?
+      end
+
+      def description(description = nil)
+        @description ||= description
+      end
+
+      def desc(desc = nil)
+        Inferno.logger.warn "'desc' has been deprecated. Use 'description' instead. Called from #{caller(0..0).first}"
```
arscan:
```suggestion
        Inferno.logger.warn "Use of 'desc' in test metadata has been deprecated. Use 'description' instead. Called from #{caller(1..1).first}"
```

PR #304, opened 2019-09-26
FI-380 Fix Typo

radamson:
Fixes a typo of the word `application` in tests that check for JSON support in the CapabilityStatement.

Right now we have some [tests for the Sequences like the Argonaut Conformance Sequence](https://github.com/onc-healthit/inferno/blob/development/test/sequence/conformance_test.rb), but without good granularity for testing this single issue nicely.

@Jammjammjamm (and others) I'm happy to add a test if folks think it's warranted, otherwise it might be easier to wait and add a better test when the individual tests are more testable seeing how small a change this is.

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


czh-orz:
I think the formats array should probably be a defined in CapabilityStatementSequence as a constant so it doesn't have to live in three different files.

Jammjammjamm:
Maybe called `JSON_FORMATS` in xml tests are added at some point.

radamson:
@czh-orz I added a method in CapabilityStatementSequence (and overrode it in the argonaut DSTU2 Conformance test) for retrieving the JSON formats instead of using a constant.  The acceptable format for DSTU2,`application/json+fhir`, changed to `application/fhir+json` in STU3 and up.

[R4 MIME-Type](https://www.hl7.org/fhir/json.html)
[DSTU2 MIME-Type](https://hl7.org/fhir/DSTU2/json.html)

arscan:
We expose provide format information in fhir_client, however, we provide the technically correct form and not the other 'allowable' forms: https://github.com/fhir-crucible/fhir_client/blob/master/lib/fhir_client/resource_format.rb

https://hl7.org/fhir/R4/http.html#mime-type

By my reading, if we say we Accept application/fhir+json, they can't shoot application/json back at us.  Should we be more strict here? 😈

radamson:
Reading through it I agree and do think we should be more strict here.

`The correct mime type SHALL be used by clients and servers.`

arscan:
@radamson what do you want to do here.  There are two unresolved questions:
1) Do we want to use the constants from fhir_client?  And perhaps add a method here or there that lets you pass the format and optional version (I view pre-r4 a special case) and have it return a string? Or is just duplicating that here an acceptable path forward because you don't view that as part of the fhir_client's 'API'?
2) Do we want to clamp down on the requirement here for allowed string formats to be returned (since we will be sending the right Accept format)?

radamson:
1. For now I'm leaning towards explicitly stating the mime-type in Inferno and not getting it from fhir_client.  I have no strong opinion on it yet, but opening up another PR in that library and releasing a new gem to get a method for accessing it seems like it should be another PR so we can get this fix in.  Plus I think having it here in Inferno makes clearer the mime-type we're looking for.
2. Yes let's clamp down.

I'll make the changes and give everyone time to review before merging.

radamson:
Updating understanding and outcome after reading through the spec again and discussing with @arscan.

No longer going to clamp down the check for the single MIME-type.

[The requirement we were referencing](https://hl7.org/fhir/R4/http.html#mime-type) appears to only apply to the Content-Type HTTP header ([which may not be required in general](https://tools.ietf.org/html/rfc7231#section-3.1.1.5)).

The test in question which was being updated is only looking at the format defined in the [Capability Statement, which allows for any MIME-Type, including just `json`](https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.format).

[The US Core IG Capability Statement also just uses `json`](https://build.fhir.org/ig/HL7/US-Core-R4/CapabilityStatement-us-core-server.json.html).

radamson:
Updated test descriptions:

**US Core**:
Screen Shot 2019-10-01 at 10 40 24 AM

**Bulk Data**:
Screen Shot 2019-10-01 at 10 39 25 AM

lib/app/modules/bulk_data/bulk_data_capability_statement_sequence.rb
```diff
@@ -65,32 +65,29 @@ def assert_operation(op_name); end
...
-            ```
-            [http://hl7.org/fhir/us/core/2019Jan/CapabilityStatement-us-core-server.html](http://hl7.org/fhir/us/core/2019Jan/CapabilityStatement-us-core-server.html)
+            [```The US Core Server **SHALL** Support json source formats for all US Core interactions.```](https://build.fhir.org/ig/HL7/US-Core-R4/CapabilityStatement-us-core-server.html)
 
-            The FHIR capability interaction require servers to describe which formats are available for clients to use.  The server must
-            explicitly state that JSON is supported. This is located in the [format element](https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.format)
+            The FHIR capability interaction require servers to describe which formats are available for clients to use.
+            The server must explicitly state that JSON is supported.  This is located in the
+            [format element](https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.format)
             of the Capability Resource.
```
Jammjammjamm:
```suggestion
            of the CapabilityStatement Resource.
```
There is no Capability resource.

lib/app/modules/bulk_data/bulk_data_capability_statement_sequence.rb
```diff
@@ -65,32 +65,29 @@ def assert_operation(op_name); end
...
-            This test checks that one of the following values are located in the [format field](https://www.hl7.org/fhir/DSTU2/json.html).
+            This test checks that one of the following values are located in the format field.
 
             * json
             * application/json
-            * application/json+fhir
+            * application/fhir+json
 
-            Note that FHIR changed the FHIR-specific JSON mime type to `application/fhir+json` in later versions of the specification.
+            This test supports the [Bulk Data US Core Data for Interoperability Use Case](https://build.fhir.org/ig/HL7/bulk-data/#us-core-data-for-interoperability)
```
Jammjammjamm: Is this requirement actually stated somewhere in that IG that we can link to, or is it being inferred?

radamson: I didn't really see a hard requirement, just the use case and was attempting to update the bulk data details as best I could.

@yunwwang any thoughts here or should this test be removed

radamson: Also happy to remove these updates if they need more discussion so that the original typo fix can get in.

radamson: removed the bulk data IG reference.

PR #305, opened 2019-09-26
FI-365 Save resources from all pages from search

czh-orz:
Currently, sequences only saves the resources returned in the first page of the search response. This PR updates each sequences to go through all the pages. These results aren't being used for anything special right now, but will allow us to build more robust tests in the future.

You can test this by looking at the HTTP requests from searches. HSPC has 10 resources per page so do a search with >10 resources. I used encounter for Adams, Daniel X in the HSPC sandbox.

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-365

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -159,14 +159,7 @@ module Inferno
               skip 'No resources appear to be available for this patient. Please use patients with more information.' unless @resources_found
 
               @#{sequence[:resource].downcase} = reply&.resource&.entry&.first&.resource
-              @#{sequence[:resource].downcase}_ary = reply&.resource&.entry&.map { |entry| entry&.resource }
-              page_count = 1
-              next_bundle = reply&.resource&.next_bundle
-              until next_bundle.nil? || page_count == 100
-                @#{sequence[:resource].downcase}_ary += next_bundle&.entry&.map { |entry| entry&.resource }
-                next_bundle = next_bundle.next_bundle
-                page_count += 1
-              end
+              @#{sequence[:resource].downcase}_ary = fetch_all_search_results(reply&.resource)
               save_resource_ids_in_bundle(#{save_resource_ids_in_bundle_arguments})
               save_delayed_sequence_references(@#{sequence[:resource].downcase})
               validate_search_reply(versioned_resource_class('#{sequence[:resource]}'), reply, search_params))
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -159,7 +159,7 @@ module Inferno
               skip 'No resources appear to be available for this patient. Please use patients with more information.' unless @resources_found
 
               @#{sequence[:resource].downcase} = reply&.resource&.entry&.first&.resource
-              @#{sequence[:resource].downcase}_ary = fetch_all_search_results(reply&.resource)
+              @#{sequence[:resource].downcase}_ary = fetch_all_bundled_resources(reply&.resource)
               save_resource_ids_in_bundle(#{save_resource_ids_in_bundle_arguments})
               save_delayed_sequence_references(@#{sequence[:resource].downcase})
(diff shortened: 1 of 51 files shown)


generator/uscore/uscore_generator.rb
```diff
@@ -158,8 +158,15 @@ def create_search_test(sequence, search_param)
 
               skip 'No resources appear to be available for this patient. Please use patients with more information.' unless @resources_found
 
-              @#{sequence[:resource].downcase} = reply.try(:resource).try(:entry).try(:first).try(:resource)
+              @#{sequence[:resource].downcase} = reply&.resource&.entry&.first&.resource
               @#{sequence[:resource].downcase}_ary = reply&.resource&.entry&.map { |entry| entry&.resource }
+              page_count = 1
```
Jammjammjamm: I think this whole thing should be a method (unfortunately it would live in `SequenceBase`, so that this code isn't duplicated in each sequence. Something like:
```
@#{sequence[:resource].downcase}_ary = fetch_all_search_results(reply&.resource)
```
That would also allow this method to be unit tested, which I think we need since we've discussed error handling around this functionality as something of particular importance.

generator/uscore/uscore_generator.rb
```diff
@@ -158,8 +158,15 @@ def create_search_test(sequence, search_param)
...
+              next_bundle = reply&.resource&.next_bundle
+              until next_bundle.nil? || page_count == 100
```
Jammjammjamm: I think 100 is pretty high. I feel like 500ms is a fairly typical response time for a FHIR search, which would mean potentially 50 seconds per sequence just on searching.

generator/uscore/uscore_generator.rb
czh-orz: I moved it to SequenceBase, but I'm not sure what the tests would look like for it. Should I try something where next_bundle never returns or maybe returns an error? Not really sure how to set that up.

generator/uscore/uscore_generator.rb
czh-orz: Good point. I changed it to 20 since that seems like a more reasonable amount.

generator/uscore/uscore_generator.rb
Jammjammjamm: Let's get together in person and go over it.

lib/app/sequence_base.rb
```diff
@@ -807,12 +807,24 @@ def date_comparator_value(comparator, date)
         case comparator
         when 'lt', 'le'
           comparator + (DateTime.xmlschema(date) + 1).xmlschema
-        when 'gt', 'ge'
+        when 'gt', 'ge'
```
arscan:
```suggestion
        when 'gt', 'ge'
```

lib/app/sequence_base.rb
```diff
@@ -813,6 +813,22 @@ def date_comparator_value(comparator, date)
           ''
         end
       end
+
+      def fetch_all_search_results(bundle)
```
Jammjammjamm: I think we discussed renaming this something like `fetch_all_bundle_pages`

lib/app/sequence_base.rb
```diff
@@ -813,6 +813,22 @@ def date_comparator_value(comparator, date)
...
+        resources = bundle&.entry&.map { |entry| entry&.resource }
+        page_count = 1
+        next_bundle_link = bundle&.link&.find { |link| link.relation == 'next' }&.url
+        next_bundle = bundle.next_bundle
```
Jammjammjamm: The code duplication in this method should be eliminated.

lib/app/sequence_base.rb
czh-orz: I changed it to fetch_all_bundled_resources

lib/app/sequence_base.rb
czh-orz: Removed duplicated code.

PR #313, opened 2019-10-03
FI-330: Make sequences more testable

Jammjammjamm:
This branch adds methods to find and execute a single test within a sequence. It also adds sample unit tests for a sequence.

The sample sequence was updated to use constants for the test names to avoid duplicating these, but this is a change we may want to revert.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-330

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/modules/smart/token_refresh_sequence.rb
+++ b/lib/app/modules/smart/token_refresh_sequence.rb
@@ -56,7 +56,8 @@ module Inferno
         LoggedRestClient.post(@instance.oauth_token_endpoint, oauth2_params, oauth2_headers)
       end
 
-      test 'Refresh token exchange fails when provided invalid Refresh Token.' do
+      INVALID_REFRESH_TOKEN_TEST = 'Refresh token exchange fails when provided invalid Refresh Token.'
+      test INVALID_REFRESH_TOKEN_TEST do
         metadata do
           id '01'
           link 'https://tools.ietf.org/html/rfc6749'
@@ -68,7 +69,8 @@ module Inferno
         assert_response_bad_or_unauthorized token_response
       end
 
-      test 'Refresh token exchange fails when provided invalid Client ID.' do
+      INVALID_CLIENT_ID_TEST = 'Refresh token exchange fails when provided invalid Client ID.'
+      test INVALID_CLIENT_ID_TEST do
         metadata do
           id '02'
           link 'https://tools.ietf.org/html/rfc6749'
@@ -144,7 +146,9 @@ module Inferno
         end
       end
 
-      test 'Server successfully refreshes the access token when optional scope parameter omitted.' do
+      REFRESH_WITHOUT_SCOPE_PARAMETER_TEST =
+        'Server successfully refreshes the access token when optional scope parameter omitted.'
+      test REFRESH_WITHOUT_SCOPE_PARAMETER_TEST do
         metadata do
           id '03'
           link 'https://tools.ietf.org/html/rfc6749'
@@ -168,7 +172,9 @@ module Inferno
         validate_and_save_refresh_response(token_response)
       end
 
-      test 'Server successfully refreshes the access token when optional scope parameter provided.' do
+      REFRESH_WITH_SCOPE_PARAMETER_TEST =
+        'Server successfully refreshes the access token when optional scope parameter provided.'
+      test REFRESH_WITH_SCOPE_PARAMETER_TEST do
         metadata do
           id '04'
           link 'https://tools.ietf.org/html/rfc6749'
--- a/lib/app/modules/smart/token_refresh_sequence.rb
+++ b/lib/app/modules/smart/token_refresh_sequence.rb
@@ -34,32 +34,10 @@ module Inferno
       INVALID_CLIENT_ID = 'INVALID_CLIENT_ID'
       INVALID_REFRESH_TOKEN = 'INVALID_REFRESH_TOKEN'
 
-      def encoded_secret(client_id, client_secret)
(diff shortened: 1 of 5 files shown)


test/sequence/smart/token_refresh_sequence_test.rb
```diff
@@ -0,0 +1,213 @@
+# frozen_string_literal: true
+
+require_relative '../../test_helper'
+
+describe Inferno::Sequence::TokenRefreshSequence do
```
Jammjammjamm: Compare to this more integration-style test which requires that the entire sequence be run for each test and depends on inspecting a `TestResult` object to check the outcome: https://github.com/onc-healthit/inferno/blob/master/test/sequence/token_refresh_sequence_test.rb

hadleynet: This looks good. Possible to DRY out this repeated block:

`before do
  @test = SEQUENCE[SOME_TEST_ID]
  @sequence = SEQUENCE.new(@instance, @client)
end`

lib/app/modules/smart/token_refresh_sequence.rb
```diff
@@ -144,7 +146,9 @@ def validate_and_save_refresh_response(token_response)
         end
       end
 
-      test 'Server successfully refreshes the access token when optional scope parameter omitted.' do
+      REFRESH_WITHOUT_SCOPE_PARAMETER_TEST =
+        'Server successfully refreshes the access token when optional scope parameter omitted.'
+      test REFRESH_WITHOUT_SCOPE_PARAMETER_TEST do
```
arscan: This is interesting, but it seems like we are encouraging a pattern of using a test "key" as well as a "name"/"short description" but not enforcing it.  If we feel strongly this is the right approach, then we could update the DSL to make a it a little easier.

```ruby
test :refresh_without_scope_parameter
   metadata do
      name 'Server successfully refreshes the access token when optional scope parameter ommited.'
      id '03'
      ...
   end
   ....
end
```

This is a little more of a burden for the test writer, because you need to come up with two good "names", instead of just one.  And you need to make sure they stay in sync if you change the name (though I would hope that the key is succint enough that it wouldn't ever change unless in the case of a major refactor, while I could imagine the longer "name" could be rephrased).

Also this seems a bit redundant with the 'test id', which is enforced to be a number in a certain format.  Though we acknowledge that test ids are much more likely to change while tests are *under development*, if you add a test in the middle then all the test ids change.  Note that after a test procedure is finalized they can't change because that defeats their purpose, at that point we'll need to allow things like '03b' if we want to add another test in the middle, or change the test substantially.

Another option is to just keep this dsl the same, and in our self-tests just reference the full name.  But if we tweak the wording we'll have to update the self-test...

Jammjammjamm: I like this suggestion. Using a symbol rather than a string signals that it's an identifier, `BIG_CONSTANTS_LIKE_THIS` are annoying, and this separates the identification of a test from display concerns.

It is a bit redundant with the id, but I think the current ids work well for the UI where you have `SD-02: Configuration from well-known endpoint contains required fields`, but work less well for tests where you have to look at the sequence to know which test`Sequence['SD-02']` really is (in addition to the reordering problem).

Jammjammjamm:
Modifications have been made based on discussions w/Rob:

`SequenceBase.test` can now take a string or symbol. If the argument is a symbol, it will be the key used to look up the test, and the `name` must be set in the `metadata` block. If a string is used as the argument, the string will be the `name`, and you won't be able to lookup the test. This is done to be backwards-compatible.
```ruby
# Current style
# Works, but you can't look up the test individually, just as things are now
test 'Refresh token exchange fails when provided invalid Refresh Token.' do
  ...
end

# New style
# Test can be looked up by the symbol passed to `test`
test :invalid_refresh_token do
  metadata do
    name 'Refresh token exchange fails when provided invalid Refresh Token.'
    ...
  end
  ...
end

# Fetch the test
TokenRefreshSequence[:invalid_refresh_token]
```

Advantages of this approach:
- Clear separation of presentation/identification concerns: `name` is strictly for presentation, `key` is strictly for identification. The data types reinforce this: string for presentation, symbol for identification.
- These keys are less likely to change than test names or test ids
- Inferno end users who want to add their own sequences can just use names and only have to worry about adding keys if they want to unit test the tests.
- You have some idea which test you're getting when looking at the code, while you wouldn't if you were looking up via the existing ids:
  `TokenRefreshSequence[:invalid_refresh_token]` vs.`TokenRefreshSequence['TR-01']`
- We can take an incremental approach to updating the sequences to use this new style

arscan:
Very cool!

I suppose the one other option is to keep the name in the same spot and just add 'key' to the metadata.  The pro to that is you can start without the key, and add it later without moving things around.  The con is that we will have an 'id' next to a 'key' which could get confusing.

Love the self tests, this is going to be a huge improvement.  @hadleynet may have an opinion on this.  Would it be worth having some kind of pattern so we can extract how well we are testing each test?  For example, `describe ':test_name' do`, or `describe '!test_name'`, or `describe SEQUENCE[:test_name] do`?  And then pull out "succeed" or "fail" to start to then be able to generate a chart of how many positive/negative things we are checking?  Or would just having this be checked on reviews be good enough?

Jammjammjamm:
I discussed this with Marc yesterday, and I'm going to attempt to do some metaprogramming with `assert` to track passes/failures for each `assert` statement.

PR #314, opened 2019-10-03
Fi 339 type filter

yunwwang:
This pull request contains changes for
* add export test with _type search parameters
* add delete request test
* add cancel request test

**Submitter:**
- [*] This pull request describes why these changes were made
- [*] Internal ticket for this PR:
FI-340 Add tests for _type parameter
- [*] Internal ticket links to this PR
https://oncprojectracking.healthit.gov/support/browse/FI-340
- [*] Internal ticket is properly labeled (Community/Program)
- [*] Internal ticket has a justification for its Community/Program label
- [*] Code diff has been reviewed for extraneous/missing code
- [*] Tests are included and test edge cases
- [*] Tests/code quality metrics have been run locally and pass

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -14,8 +14,26 @@ class BulkDataPatientExportSequence < SequenceBase
       requires :token
       conformance_supports :Patient
 
-      def check_export_kick_off(klass)
-        reply = export_kick_off(klass)
+      attr_writer :run_all_kick_off_tests
```
Jammjammjamm: It's safer to use an `attr_accessor` and only reference the accessors rather than the instance variable itself.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -14,8 +14,26 @@ class BulkDataPatientExportSequence < SequenceBase
...
+
+      def check_export_kick_off(klass, search_params: nil)
+        @search_params = search_params
+
+        reply = export_kick_off(klass, search_params: search_params)
+
+        # Servers unable to support _type SHOULD return an error and OperationOutcome resource
+        # so clients can re-submit a request omitting the _type parameter.
+        if search_params.present? && reply.code > 400
```
Jammjammjamm: Should this be `>=`? Same thing in line 118, so maybe this should be a little helper method.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -14,8 +14,26 @@ class BulkDataPatientExportSequence < SequenceBase
...
+          response_body = JSON.parse(reply.body)
+          message = ''
+          response_body['issue'].each do |issue|
+            message += issue['diagnostics'].presence || ''
```
Jammjammjamm: Does some whitespace need to be added between messages? If so this might be easiest:
```
message = response_body['issue']
  .map { |issue| issue['diagnostics'].presence }
  .compact
  .join(' ') // or whatever whitespace is appropriate
```

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -14,8 +14,26 @@ class BulkDataPatientExportSequence < SequenceBase
...
+          end
+
+          skip message
+        end
+
+        @server_support_type_parameter = search_params.present?
```
Jammjammjamm: How does the presence of any search param indicate that the server supports the _type parameter? Line 26 also seems to be assuming that `_type` is the only possible search param.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -57,13 +75,18 @@ def check_export_status(url = @content_location, timeout: 180)
         assert_output_has_type_url
       end
 
-      def assert_output_has_type_url(output = @output)
+      def assert_output_has_type_url(output = @output,
+                                     search_params = @search_params)
         skip 'Sever response did not have output data' unless output.present?
 
+        search_type = search_params['_type'].split(',').map(&:strip) if search_params.present? && search_params.key?('_type')
```
Jammjammjamm:
```suggestion
        search_type = search_params['_type'].split(',').map(&:strip) if search_params&.key?('_type')
```

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -90,6 +113,18 @@ def check_ndjson(ndjson, type)
         end
       end
 
+      def check_delete_request(url)
+        reply = @client.delete(url, {})
+        skip 'Server did not accept client request to delete export file after export completed' if reply.code > 400
```
Jammjammjamm: Is the skip really needed? The test is marked `optional`.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
yunwwang: It is true that _type is the only parameter for now. Other parameters will be add later. Still worth to fix it.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
yunwwang: Yes since "not accept client delete request" is also a valid response. This test is focus on the return code after server accepted client's request. Test 08 is already marked as optional. Will mark Test 09 as optional too.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -115,44 +150,55 @@ def check_ndjson(ndjson, type)
...
+
+      test 'Server shall return "202 Accepted" and "Content-location" for $export operation' do
+        metadata do
+          id '03'
           link 'https://build.fhir.org/ig/HL7/bulk-data/export/index.html#bulk-data-kick-off-request'
           desc %(
           )
         end
 
+        skip 'Skip testing $export without parameters' if @server_support_type_parameter && !@run_all_kick_off_tests
```
Jammjammjamm:
```suggestion
        skip 'Skip testing $export without parameters' if @server_support_type_parameter && !run_all_kick_off_tests
```
Now that you have an `attr_accessor`, use it here. This protects against typos in instance variable names. If someone typed `@run_all_kickoff_tests` instead of `@run_all_kick_off_tests`, they'd get `nil` and the tests would keep running and they may or may not notice that this test is no longer working as intended. A typo in the name of an accessor, however, would raise an exception so that the error could be discovered immediately.

lib/app/modules/bulk_data/bulk_data_patient_export_sequence.rb
```diff
@@ -14,8 +14,26 @@ class BulkDataPatientExportSequence < SequenceBase
...
+          response_body = JSON.parse(reply.body)
+          message = response_body['issue']
+            .map { |issue| issue['diagnostics'].presence }
+            .compact
+            .join(' ')
+
+          skip message
+        end
+
+        @server_support_type_parameter = search_params&.key?('_type')
```
Jammjammjamm: I think this should be named `@server_supports_type_parameter`. Also, it should be moved above line 26 so that it can be used in the condition rather than repeating `search_params&.key?('_type')`

PR #317, opened 2019-10-08
FI-385: Check for US Core profile support

Jammjammjamm:
Update the US Core Capability Statement sequence to include a test that checks for support of the US Core Patient profile + one other US Core profile.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-385

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/modules/us_core_guidance/us_core_r4_capability_statement_sequence.rb
+++ b/lib/app/modules/us_core_guidance/us_core_r4_capability_statement_sequence.rb
@@ -88,9 +88,10 @@ module Inferno
         'Provenance' => ['http://hl7.org/fhir/us/core/StructureDefinition/us-core-provenance']
       }.freeze
 
-      test 'FHIR server capability states JSON support' do
+      test :json_support do
         metadata do
           id '04'
+          name 'FHIR server capability states JSON support'
           link 'http://hl7.org/fhir/us/core/2019Jan/CapabilityStatement-us-core-server.html'
           description %(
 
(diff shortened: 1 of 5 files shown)


arscan:
Is there a requirement somewhere that the server SHALL list all supported profiles in their capability statement if they support it?

arscan:
I asked the question here: https://chat.fhir.org/#narrow/stream/179175-argonaut/topic/How.20do.20servers.20'claim.20conformance'.20to.20a.20profile.3F

The closest we could find right now is that the core specification says that a server SHOULD fill out that information, but it is not a requirement.  If we are given no more direction, we should either turn this into an optional test, or spit out a warning, and reference that should statement in the core spec: 

> Servers SHALL specify what resource types and operations are supported, and SHOULD also specify profiles for each resource type.  
> https://www.hl7.org/fhir/capabilitystatement.html#notes

arscan:
This isn't backed as a requirement in US Core. What we could do is just check for RESOURCE level support here instead, as that is required.

arscan:
And throw a warnings if they don't explicitely state profile support within the resource, as that is a SHOULD.

lib/app/modules/us_core_guidance/us_core_r4_capability_statement_sequence.rb
```diff
@@ -55,9 +55,43 @@ class UsCoreR4CapabilityStatementSequence < CapabilityStatementSequence
         * [DSTU2 Conformance Statement](https://www.hl7.org/fhir/DSTU2/conformance.html)
       )
 
-      test 'FHIR server capability states JSON support' do
+      PROFILES = {
```
arscan: This is a little bit redundant with the data provided in app/helpers/validation.rb, and perhaps in other places.  I'm a little worried if these change we won't remember to update all the different places that these get referenced.  But this is a bigger problem then just this PR, so I'm good with leaving this as is.

Jammjammjamm: Yeah, we might want to come up with some general module-level structure for storing these kinds of module-wide knowledge/behaviors.

PR #318, opened 2019-10-10
FI-232: valueset display

okeefm:
This PR adds the ability to see which valuesets/validators are loaded into an Inferno server from the UI, rather than having to go into the console. To get to the modal that displays the valuesets, click on the "version" link in the bottom right hand corner. Eventually this link will contain additional server status info (stuff that's in the config files now), but this is a first step.

Screenshot:

Screen Shot 2019-10-10 at 11 23 05 AM

Screenshot when there are no valuesets in Inferno:

Screen Shot 2019-10-11 at 7 28 09 AM

NOTE: IF #234 gets merged before this one, this will require a rebase.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-232

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


radamson:
Can you add a screen shot of how it looks to the PR?

lib/app/utils/terminology.rb
```diff
@@ -28,6 +28,8 @@ class Terminology
 
     @known_valuesets = {}
 
+    @@loaded_validators = {}
```
radamson: why class variable instead of instance variable?

I think you should use an instance variable here similar to `@known_valuesets`

lib/app/utils/terminology.rb
```diff
@@ -220,13 +222,18 @@ def self.load_validators(directory = 'resources/terminology/validators/bloom')
...
         FHIR::DSTU2::StructureDefinition.validates_vs(validator[:url], &validate_fn)
+        @@loaded_validators[validator[:url]] = validator[:count]
       end
     end
 
     def self.get_valueset(url)
       @known_valuesets[url].valueset || raise(UnknownValueSetException, url)
     end
 
+    def self.loaded_validators
```
radamson: could make this an `attr_reader` with an instance variable.

lib/app/views/server_state.erb
```diff
@@ -0,0 +1,19 @@
+<div id="server-state">
+  <h4>Valuesets Used in Inferno:</h4>
+  <table class="table table-striped">
+    <thead>
+      <tr>
+        <th>Valueset URL</th>
+        <th>Code count</th>
```
radamson: Should this be `Concept Count`.

That seems to be how they are normally referred to:

e.g.
https://www.hl7.org/fhir/valueset-patient-contactrelationship.html#expansion
https://www.hl7.org/fhir/us/core/ValueSet-us-core-careteam-provider-roles.html#expansion

lib/app/utils/terminology.rb
okeefm: See above.

lib/app/utils/terminology.rb
Jammjammjamm: In that case I'd recommend using a class instance variable rather than a class variable. If you refer to an instance variable within a class method, you get a class instance variable:
https://github.com/onc-healthit/inferno/blob/development/lib/app/sequence_base.rb#L311

They're basically like class variables without the weirdness around inheritance.

radamson: Thats sort of what is being done with `@known_valuesets`, minus the method that exposes it.

Jammjammjamm: Ugh, given the current prevalance of class variables here I wonder whether it's really worth changing it. If we really care about avoiding class variables, I think the thing to probably do is wrap pretty much the entire class body in a `class << self` block and replace all the  class variables with instance variables. Then you could use attr_reader/writers, too. Don't know that it makes sense to undertake that as part of this ticket, though.

radamson: I wouldn't remove the other ones, just make this one a class instance variable.

This other [PR](https://github.com/onc-healthit/inferno/pull/234) removes them all from the class already.  I just don't want to add another.

radamson: unless you don't think the class instance variables are any better.

Jammjammjamm: Since the other PR removes them, it makes sense for this to be a class instance variable.

okeefm: Sure, that works for me.

okeefm:
Updates made to use a class instance variable, and to add a dialog if there are no valuesets in the server. Screenshot above for the latter.

arscan:
@okeefm or @radamson could you address the conflict now that I've merged in the R4 terminology update?

okeefm:
@arscan Rebased

arscan:
I made some updates to the wording based on what is technically here.

PR #343, opened 2019-11-04
Fi 444 try multiple values in search

czh-orz:
This will make the generator look for fixed values and value set info when creating the first search of a sequence. The first search will search on all the possible values until it finds something. This also updates the capability statement to include new diagnostic_report searches.

Here are the list of sequences that this applies to:
Careplan
Careteam
Smoking Status
Lab Results
Pediatric Weight Height
Pediatric BMI Age
DiagnosticReport Lab	
DiagnosticReport Note
Pulse Oximetry

You can verify that this is working by running one of the above sequences and looking at the requests for the first search. You can use patient SMART-1288992 on https://api.logicahealth.org/czhour4/open. There should now be a MedicationRequest with an intent of 'Order'. You should see three MedicationRequest requests before it hits the order search.

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-444

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/app/modules/uscore_v3.0.1/pediatric_bmi_for_age_sequence.rb
```diff
@@ -76,22 +74,26 @@ def validate_resource_item(resource, property, value)
           versions :r4
         end
 
-        search_params = { patient: @instance.patient_id, code: '59576-9' }
+        code_val = ['59576-9']
+        code_val.each do |val|
```
Jammjammjamm: Could this be `codes.each do |code|` instead?

lib/app/modules/uscore_v3.0.1/us_core_smokingstatus_sequence.rb
```diff
@@ -177,9 +179,9 @@ def validate_resource_item(resource, property, value)
         assert !@observation.nil?, 'Expected valid Observation resource to be present'
 
         patient_val = @instance.patient_id
-        code_val = get_value_for_search_param(resolve_element_from_path(@observation_ary, 'code'))
-        date_val = get_value_for_search_param(resolve_element_from_path(@observation_ary, 'effectiveDateTime'))
-        search_params = { 'patient': patient_val, 'code': code_val, 'date': date_val }
+        category_val = get_value_for_search_param(resolve_element_from_path(@observation_ary, 'category'))
+        status_val = get_value_for_search_param(resolve_element_from_path(@observation_ary, 'status'))
```
Jammjammjamm: Why not `category` and `status` rather than `category_val` and `status_val`?

arscan:
Could you give a patient id and server and resource that shows this running?

I'm using patientid `SMART-1288992` on your open server but that doesn't have MedicationRequests, which seems to be what this applies to.

lib/app/modules/uscore_v3.0.1/us_core_smokingstatus_sequence.rb
czh-orz: I think I did that because encounter has a search by class+patient and class is a protected keyword.

lib/app/modules/uscore_v3.0.1/pediatric_bmi_for_age_sequence.rb
czh-orz: It's the same issue as the question below. If the parameter is  'class', then we run into issues. This isn't an issue in US core, but I figured the awkward naming was worth making it more generic.

generator/uscore/metadata_extractor.rb
```diff
@@ -265,6 +282,15 @@ def add_special_cases(metadata)
             sequence[:searches].unshift(category_search)
           end
         end
+
+        # search by patient + intent first for medication request sequence
+        medication_request_sequence = metadata[:sequences].find { |sequence| sequence[:resource] == 'MedicationRequest' }
+        intent_search = medication_request_sequence[:searches].find { |param| param[:names] == ['patient', 'intent'] }
+        unless intent_search.nil?
```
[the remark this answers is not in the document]
czh-orz: I think I had this because 3.0.0 didn't have a search by intent for medicationrequest.

generator/uscore/metadata_extractor.rb
```diff
@@ -265,6 +282,15 @@ def add_special_cases(metadata)
             sequence[:searches].unshift(category_search)
           end
         end
+
+        # search by patient + intent first for medication request sequence
```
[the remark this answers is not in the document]
czh-orz: Moved to another function

generator/uscore/metadata_extractor.rb
```diff
@@ -206,11 +206,30 @@ def add_search_param_descriptions(profile_definition, sequence)
...
             param_metadata[:type] = profile_element['type'].first['code']
             param_metadata[:contains_multiple] = (profile_element['max'] == '*')
+            if param_metadata[:contains_multiple]
+              slices = profile_definition['snapshot']['element'].select { |el| el['path'] == path && el['sliceName'] }
+              slices.each do |slice|
+                param_metadata[:values] << slice['patternCodeableConcept']['coding'].first['code'] if slice['patternCodeableConcept']
+              end
+            elsif param_metadata[:type] == 'CodeableConcept'
+              fixed_code_els = profile_definition['snapshot']['element'].select { |el| el['path'] == "#{profile_element['path']}.coding.code" && el['fixedCode'].present? }
+              param_metadata[:values] += fixed_code_els.map { |el| el['fixedCode'] } if fixed_code_els.any?
```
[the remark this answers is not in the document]
czh-orz: fixed

generator/uscore/metadata_extractor.rb
```diff
@@ -206,11 +206,30 @@ def add_search_param_descriptions(profile_definition, sequence)
...
           param_metadata = {
             path: path,
-            comparators: {}
+            comparators: {},
+            values: []
           }
           if !profile_element.nil?
             param_metadata[:type] = profile_element['type'].first['code']
             param_metadata[:contains_multiple] = (profile_element['max'] == '*')
+            if param_metadata[:contains_multiple]
```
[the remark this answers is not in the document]
czh-orz: I moved it to another function. It checks a lot of different stuff so not sure how to make simpler than it currently is.

czh-orz:
Updated the PR description with more testing info.

generator/uscore/metadata_extractor.rb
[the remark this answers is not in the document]
Jammjammjamm: Wherever there is a comment explaining what the next few lines do, it could instead be a method whose name says what it does.
```
if param_metadata[:contains_multiple]
  add_values_from_slices(args)
elsif param_metadata[:type] == 'CodeableConcept'
  add_values_from_codeable_concepts(args)
end
add_whatever_this_one_is(args)
if value_set_binding
  add_values_from_valueset(args)
end
add_values_from_metadata unless ...
```

generator/uscore/metadata_extractor.rb
```diff
@@ -226,6 +228,30 @@ def add_search_param_descriptions(profile_definition, sequence)
...
+        param_metadata[:values] << profile_element['patternCodeableConcept']['coding'].first['code'] if profile_element['patternCodeableConcept']
+        # look for values in the valueset the element is bound to
+        valueset_binding = profile_element['binding']
+        if valueset_binding
+          value_set = resources_by_type['ValueSet'].find { |res| res['url'] == valueset_binding['valueSet'] }
+          codes = value_set['compose']['include'].reject { |code| code['concept'].nil? } if value_set.present?
+          param_metadata[:values] += codes.map { |code| code['concept'].first['code'] } if codes.present?
+        end
+        # if none found so far, add all the valid codes from metadata (less strict)
+        param_metadata[:values] = fhir_metadata['valid_codes'].values.flatten unless param_metadata[:values].any? || fhir_metadata.nil? || fhir_metadata['valid_codes'].nil?
```
Jammjammjamm: Use `.present?/.blank?` rather than `.any?` or `.nil?`. `.any?` isn't nil-safe, and `.nil?` returns true for empty arrays, hashes, and strings, which usually isn't what you want.

czh-orz: Fixed

generator/uscore/metadata_extractor.rb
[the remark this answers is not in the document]
czh-orz: Fixed

Jammjammjamm:
I think this is good once it's rebased.

generator/uscore/metadata_extractor.rb
```diff
@@ -226,6 +228,47 @@ def add_search_param_descriptions(profile_definition, sequence)
...
 
+      def add_valid_codes(profile_definition, profile_element, fhir_metadata, param_metadata, path)
+        if param_metadata[:contains_multiple]
+          add_values_from_slices(param_metadata, profile_definition, path)
+        elsif param_metadata[:type] == 'CodeableConcept'
+          add_values_from_fixed_codes(param_metadata, profile_definition, profile_element)
+          add_values_from_patterncodeableconcept(param_metadata, profile_element)
+        end
+        add_values_from_valueset_binding(param_metadata, profile_element)
+        add_values_from_resource_metadata(param_metadata, fhir_metadata)
```
radamson: if the binding on the resource is `example`, but the binding on the defining profile is `required` it looks like the resource values will still be added (even though they probably shouldn't be).

generator/uscore/metadata_extractor.rb
```diff
@@ -206,11 +206,13 @@ def add_search_param_descriptions(profile_definition, sequence)
           profile_element = profile_definition['snapshot']['element'].select { |el| el['id'] == path }.first
           param_metadata = {
             path: path,
-            comparators: {}
+            comparators: {},
+            values: []
```
radamson: Could this be a `Set` in case there are repeated codings from the different sources?

generator/uscore/uscore_generator.rb
```diff
@@ -369,6 +350,63 @@ def param_value_name(param)
...
+            skip 'No resources appear to be available for this patient. Please use patients with more information.' unless @resources_found
+
+            @#{sequence[:resource].downcase} = reply&.resource&.entry&.first&.resource
+            @#{sequence[:resource].downcase}_ary = fetch_all_bundled_resources(reply&.resource)
+            save_resource_ids_in_bundle(#{save_resource_ids_in_bundle_arguments})
+            save_delayed_sequence_references(@#{sequence[:resource].downcase}_ary)
+            validate_search_reply(versioned_resource_class('#{sequence[:resource]}'), reply, search_params)
+          )
+        else
+          # assume only patient + one other parameter
```
radamson: This method is pretty big, is it worth breaking out the consequent and alternative logic in the if statement into their own methods?  I don't want this to be a show stopper for the PR if it doesn't make sense though.

generator/uscore/metadata_extractor.rb
czh-orz: It'll only add values from the resource if it didn't find any from the profile (checks in add_values_from_resource_metadata) so I don't think we need to worry about this.

generator/uscore/metadata_extractor.rb
czh-orz: fixed

generator/uscore/uscore_generator.rb
czh-orz: Moved both to their own function.

generator/uscore/metadata_extractor.rb
radamson: ahhhh I see. based on the method name I expected it was just adding them.  I do think it would be nice to have the embedded if statement up in this method to keep them all together, but the method is so small I'm cool leaving it for now.

PR #347, opened 2019-11-14
FI-483: Generate unit tests for US Core authorization tests

Jammjammjamm:
This PR is the first step in generating unit tests for the US Core sequences. It generates unit tests for the `Server rejects *** search without authorization` US Core tests. Dynamic search parameter values are a particular challenge for generating unit tests, and for now the tests that require these depend on manually created fixtures.

The tests are generated along with the US Core sequences when `rake inferno:generate[uscore,uscore_v3.0.1]` is run.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-483
- n/a Tests are included and test edge cases

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


generator/uscore/templates/unit_tests/authorization_unit_test.rb.erb
```diff
@@ -0,0 +1,42 @@
+describe 'unauthorized search test' do
```
radamson: It would be nice to have a comment at the top indicating that this test is autogenerated.  Also a note that any changes may be overwritten when regenerating tests (I assume that would happen).

generator/uscore/templates/unit_tests/authorization_unit_test.rb.erb
```diff
@@ -0,0 +1,42 @@
...
+      .map { |resource| FHIR.from_contents(resource.to_json) }
+    @sequence.instance_variable_set(:'<%= search_param[:variable_name] %>', <%= search_param[:variable_name] %>)
+  <% end %>
+<% end %>
+    @query = {
+      <%= search_param_string %>
+    }
+  end
+
+  it 'fails when the token refresh response has a success status' do
```
radamson: I'm cool with printing this over again in every test, but I could see some tradeoffs here.  If we put this once in some sort of master/helper file and then pointed other tests here we'd know that all these tests were using the same code.  Plus it might save some lines of code.  For now it's not a big deal and I'll leave it to your discretion.

Jammjammjamm: I think we may want to optimize these tests for being simple and readable rather than being DRY. Since they're generated, I don't think repetition is really harmful, since we can still update a template in one place and have it update that bit of code everywhere.

I also didn't worry much about doing things optimally yet in this PR. This is the simplest of the tests to add, so after I add the second or third sets of unit tests the requirements and a potentially better structure (especially for the generator) will be more clear.

PR #352, opened 2019-11-18
Fi 490 set authorization search params

czh-orz:
This PR sets the search params for authorization tests back to the first search. If there are multiple possible search values, it will use the first one. 

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-490

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -147,10 +147,13 @@ module Inferno
 
         search_parameters = first_search[:names]
         search_params = if search_parameters == ['patient'] || sequence[:delayed_sequence] || search_param_constants(search_parameters, sequence)
+                          unit_test_params = get_search_param_hash(search_parameters, sequence)
                           get_search_params(search_parameters, sequence)
                         else
                           non_patient_search_param = search_parameters.find { |param| param != 'patient' }
                           non_patient_value = sequence[:search_param_descriptions][non_patient_search_param.to_sym][:values].first
+                          unit_test_params = { patient: '@instance.patient_id' }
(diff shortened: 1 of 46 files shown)


generator/uscore/uscore_generator.rb
```diff
@@ -131,10 +131,21 @@ def create_authorization_test(sequence)
         first_search = find_first_search(sequence)
         return if first_search.nil?
 
+        search_parameters = first_search[:names]
+        search_params = if search_parameters == ['patient'] || sequence[:delayed_sequence] || search_param_constants(search_parameters, sequence)
```
Jammjammjamm: I think all of the logic for what search params should be used belongs in `get_search_param_hash`. This method should be able to just call `get_search_params` and `get_search_param_hash`.

czh-orz: The problem with this is that sometimes get_search_params needs the value from resolve_element_from_path and here it needs it from the fixed code. I'd have to add a thing just for the case of the authorization test. I think I can get around this by just putting the authorization test after the search tests which I actually like. I'll try that out and put it in this branch.

Jammjammjamm: This logic only depends on `search_parameters` and `sequence`, which are the arguments which `get_search_param_hash` already takes, so it should be possible to move the logic there.

czh-orz: yep, much cleaner now.

radamson:
It looks like this has merge conflicts and needs a rebase 😞

PR #359, opened 2019-11-21
FI-491 bulk data authentication

yunwwang:
This pull request contains changes for Bulk Data Authorization sequence tests

**Submitter:**
https://oncprojectracking.healthit.gov/support/browse/FI-491

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,269 @@
...
+                    content_type: 'application/x-www-form-urlencoded',
+                    scope: 'system/*.read',
+                    grant_type: 'client_credentials',
+                    client_assertion_type: 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer',
+                    iss: @instance.bulk_client_id,
+                    sub: @instance.bulk_client_id,
+                    aud: @instance.bulk_token_endpoint,
+                    exp: 5.minutes.from_now,
+                    jti: SecureRandom.hex(32))
+        id_token = JSON::JWT.new(
```
Jammjammjamm: Maybe just call this `token`? `id_token` makes it seem like it's OIDC related.

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,269 @@
...
+          {
+            content_type: content_type,
+            accept: 'application/json'
+          }.compact
+
+        uri = Addressable::URI.new
+        uri.query_values = payload
+
+        response = LoggedRestClient.post(@instance.bulk_token_endpoint, uri.query, header)
+        response
```
Jammjammjamm: Don't need the `response` variable.

lib/modules/bulk_data/test/authorization_sequence_test.rb
```diff
@@ -0,0 +1,171 @@
...
+    @access_token = {
+      'token_type' => 'bearer',
+      'expires_in' => 900,
+      'access_token' => 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0b2tlbl90eXBlIjoiYmVhcmVyIiwiZXhwaXJlc19pbiI6OTAwLCJpYXQiOjE1NzM5NDU1MDQsImV4cCI6MTU3Mzk0NjQwNH0.Ds-9HxQPJshkPYYBowJXltTaX2T6MSv_qYnZLjteTH8',
+      'scope' => 'system/*.read'
+
+    }
+  end
+
+  def self.it_test_rquired_parameter(test_id, request_headers: nil)
```
Jammjammjamm:
```suggestion
  def self.it_tests_required_parameter(test_id, request_headers: nil)
```

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,269 @@
...
+          id '01'
+          name 'Bulk Data authorization request shall use content_type "application/x-www-form-urlencoded"'
+          link 'https://build.fhir.org/ig/HL7/bulk-data/authorization/index.html#protocol-details'
+          description %(
+            After generating an authentication JWT, the client requests a new access token via HTTP POST to the FHIR authorization server’s token endpoint URL, using content-type application/x-www-form-urlencoded
+          )
+        end
+
+        response = authorize(content_type: 'application/json')
+        assert_response_bad_or_unauthorized(response)
```
Jammjammjamm: I don' think an unauthorized response is allowed in the spec: https://tools.ietf.org/html/rfc6749#page-45

lib/modules/bulk_data/test/authorization_sequence_test.rb
```diff
@@ -0,0 +1,171 @@
...
+    before do
+      @test = @sequence_class[test_id]
+      @sequence = @sequence_class.new(@instance, @client)
+    end
+
+    it 'pass with stastus code 400' do
+      a_request = stub_request(:post, @instance.bulk_token_endpoint)
```
Jammjammjamm: When inferno is making requests, I think it is important that we test the request as well as the response. This stub doesn't verify that each test's request is malformed in the way that we expect.

lib/modules/bulk_data/test/authorization_sequence_test.rb
```diff
@@ -0,0 +1,171 @@
...
+    }
+  end
+
+  def self.it_test_rquired_parameter(test_id, request_headers: nil)
+    before do
+      @test = @sequence_class[test_id]
+      @sequence = @sequence_class.new(@instance, @client)
+    end
+
+    it 'pass with stastus code 400' do
```
Jammjammjamm:
```suggestion
    it 'passes with status code 400' do
```

lib/modules/bulk_data/test/authorization_sequence_test.rb
```diff
@@ -0,0 +1,171 @@
...
+        )
+
+      error = assert_raises(Inferno::AssertionException) do
+        @sequence.run_test(@test)
+      end
+
+      assert_match(/access_token is empty/, error.message)
+    end
+
+    it 'fail when server returns empty token_type' do
```
Jammjammjamm: We should test for empty `expires_in` and `scope` too.

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,269 @@
...
+      end
+
+      test :require_grant_type do
+        metadata do
+          id '03'
+          name 'Bulk Data authorization request shall use grand_type "client_credentials"'
+          link 'https://build.fhir.org/ig/HL7/bulk-data/authorization/index.html#protocol-details'
+          description %(
+            * grant_type	required	Fixed value: client_credentials
```
Jammjammjamm: The extra whitespace isn't rendered in the front end. Maybe markdown table formatting would work?

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,269 @@
...
+
+      test :require_system_scope do
+        metadata do
+          id '02'
+          name 'Bulk Data authorization request shall use "system" scope'
+          link 'https://build.fhir.org/ig/HL7/bulk-data/authorization/index.html#scopes'
+          description %(
+            clients SHALL use “system” scopes.
+
+            System scopes have the format system/(:resourceType|*).(read|write|*)
```
Jammjammjamm: This string doesn't display properly in the front end.

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
yunwwang: renamed to jwt_token

lib/modules/bulk_data/test/authorization_sequence_test.rb
yunwwang: Malformed requests are tested in test sequence 01-11.

Jammjammjamm: Yes, but this stub doesn't verify that the requests we send are actually malformed.

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
yunwwang: Good point. Changed to markdown table

lib/modules/bulk_data/test/authorization_sequence_test.rb
yunwwang: Can you clarify? We are not going to send malformed request in test 12.

yunwwang: Oh. I was looking at the wrong line. Because the request doesn't matter here. What we are testing here is that if the server returns 400, the test sequence should handle 400 correctly.

Jammjammjamm: Yes, but we also need to test that Inferno is making the requests correctly.

These tests basically have two parts:
- Make a request in a particular way
- Expect a particular type of response

The unit tests are currently returning particular types of responses and verifying that the test passes or fails. This is good, but it only checks half of the tests behavior.

The unit tests are not checking whether each test makes its request in the particular way it needs to. For example, one test needs to omit `grant_type`, another needs to have an invalid `iss`. If the unit tests don't check that the requests match these requirements, then at any time in the future the code could change such that the requests are no longer being made in the correct way without causing the unit tests to fail. The way to check these requests is to call `stub_request` separately for each unit test so that it will only match against the correct type of request rather than the current behavior where it matches any POST to the correct endpoint.

yunwwang: The client_claim is unique for each request since both expires_in and jti change very time a claim is created. So the claim generated from test sequence and test stub are different though parameters having same values. It is possible to inject those values into the test sequence as attribute.

Jammjammjamm: It doesn't have to be an exact match. You should be able to use a block in webmock to inspect the body and verify that the portion of the jwt that this test cares about is correct: https://github.com/bblimke/webmock#matching-requests-against-provided-block

lib/app/views/default.erb
```diff
@@ -226,7 +226,39 @@
...
+                          If you have provided your JWKS at registration time, please enter the public key here. Leave it blank if the keys have not
+                          been exchanged ahead of time.
+                        </small>
+
+              </label>
+              <textarea type="text" class="form-control" name="bulk_public_key" id="bulk_public_key" placeholder="" rows="5"><%=instance.bulk_public_key %></textarea>
+        </div>
+        <div class="form-group" data-requiredby="<%= test_set.variable_required_by(:bulk_private_key).map {|seq| seq.sequence_name}.join(',')%>" data-definedby="" data-prerequisite="bulk_private_key">
+
+              <label for="bulk_private_key">Bulk Private Key
```
radamson: This file needs to be rebased before it can be really reviewed.  @arscan added some HTML elements with similar element ids that may need deconfliction (e.g. `bulk_private_key`)

https://github.com/onc-healthit/inferno/blob/c426caeb4ae413149ff84b9218e2ed5939c1eb6c/lib/app/views/default.erb#L330-L340

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,277 @@
...
+        assert_response_bad(response)
+      end
+
+      test :require_system_scope do
+        metadata do
+          id '02'
+          name 'Bulk Data authorization request shall use "system" scope'
+          link 'https://build.fhir.org/ig/HL7/bulk-data/authorization/index.html#scopes'
+          description %(
+            clients SHALL use “system” scopes.
```
radamson:
```suggestion
            Servers SHALL reject client requests which do not include `system` scopes.
```

This test checks that servers reject client requests that do not include the system scope and the description should reflect that.

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,277 @@
...
+
+            System scopes have the format system/(:resourceType&#124;&#42;).(read&#124;write&#124;&#42;)
```
radamson:
```suggestion
            System scopes have the format `system/(:resourceType&#124;&#42;).(read&#124;write&#124;&#42;)`
```

We should format the scope format.

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,277 @@
...
+        end
+
+        response = authorize(scope: 'user/*.read')
+        assert_response_bad(response)
+      end
+
+      test :require_grant_type do
+        metadata do
+          id '03'
+          name 'Bulk Data authorization request shall use grand_type "client_credentials"'
```
radamson:
```suggestion
          name 'Bulk Data authorization request shall use grant_type "client_credentials"'
```

`grant` spelling

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,277 @@
...
+          link 'https://build.fhir.org/ig/HL7/bulk-data/authorization/index.html#protocol-details'
+          description %(
+            | --- | --- | --- |
```
radamson: Although the Markdown table helped a bit I think this needs some prose.

Screen Shot 2019-11-27 at 9 35 42 AM

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
```diff
@@ -0,0 +1,277 @@
...
+        assert_response_bad(response)
+      end
+
+      test :return_access_token do
+        metadata do
+          id '12'
+          name 'Bulk Data Token Endpoint shall return access token'
+          link 'https://build.fhir.org/ig/HL7/bulk-data/authorization/index.html#issuing-access-tokens'
+          description %(
+            If the access token request is valid and authorized, the authorization server SHALL issue an access token in response.
```
radamson: This is a good example of the description that should go with each table.

lib/modules/bulk_data/test/authorization_sequence_test.rb
yunwwang: Have solved most problem but the correct_key test is still a headache since the private key used could not be retrieved/tested from the payload

Jammjammjamm: That is for test 11? Would it be reasonable to check that the signature can't be validated with the expected public key?

lib/modules/bulk_data/bulk_data_authorization_sequence.rb
yunwwang: One side effect is that the updated name is really long.
image

lib/modules/bulk_data/test/authorization_sequence_test.rb
yunwwang: I was thinking how to test that it uses a specific private key. You are right. I only need to test if it not uses the correct key.

yunwwang: The logic is so complicated. I think I need another set of unit tests to test these unit tests

PR #364, opened 2019-11-25
FI-496 fix metadata errors

czh-orz:
This PR fixes some of the problems in the tests that were caused by metadata errors. It adds the 'ge', comparator and changes Procedure.occurenceDateTime to Procedure.performedDateTime.

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-496

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -181,7 +181,7 @@ module Inferno
               sequence[:must_supports] <<
                 {
                   type: 'element',
-                  path: path.gsub('[x]', type['code'].slice(0).capitalize + type['code'].slice(1..-1))
+                  path: path.gsub('[x]', capitalize_first_letter(type['code']))
                 }
             end
           else
@@ -197,13 +197,10 @@ module Inferno
       def add_search_param_descriptions(profile_definition, sequence)
         sequence[:search_param_descriptions].each_key do |param|
           search_param_definition = @resource_by_path[search_param_path(sequence[:resource], param.to_s)]
-          path_parts = search_param_definition['xpath'].split('/f:')
-          if param.to_s != '_id'
-            path_parts[0] = sequence[:resource]
-            path = path_parts.join('.')
-          else
-            path = path_parts[0]
-          end
+          path = search_param_definition['expression']
+          path = path.gsub(/.where\((.*)/, '')
+          as_type = path.scan(/.as\((.*?)\)/).flatten.first
+          path = path.gsub(/.as\((.*?)\)/, capitalize_first_letter(as_type)) if as_type.present?
           profile_element = profile_definition['snapshot']['element'].select { |el| el['id'] == path }.first
           param_metadata = {
             path: path,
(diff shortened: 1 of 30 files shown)


generator/uscore/metadata_extractor.rb
```diff
@@ -284,6 +285,20 @@ def add_element_definitions(profile_definition, sequence)
         end
       end
 
+      def fix_metadata_errors(metadata)
+        # Procedure's date search param definition says Procedure.occurenceDateTime even though Procedure doesn't have an occurenceDateTime
+        procedure_sequence = metadata[:sequences].find { |sequence| sequence[:resource] == 'Procedure' }
+        procedure_sequence[:search_param_descriptions][:date][:path] = 'Procedure.performedDateTime'
```
Jammjammjamm: Does this also need to support `Procedure.performedPeriod`?

generator/uscore/metadata_extractor.rb
```diff
@@ -180,7 +181,7 @@ def add_must_support_elements(profile_definition, sequence)
               sequence[:must_supports] <<
                 {
                   type: 'element',
-                  path: path.gsub('[x]', type['code'].slice(0).capitalize + type['code'].slice(1..-1))
+                  path: path.gsub('[x]', capitalize_first_letter(type['code']))
```
Jammjammjamm:
```suggestion
                  path: path.gsub('[x]', type['code'].capitalize)
```

generator/uscore/metadata_extractor.rb
```diff
@@ -196,13 +197,10 @@ def add_must_support_elements(profile_definition, sequence)
...
-          if param.to_s != '_id'
-            path_parts[0] = sequence[:resource]
-            path = path_parts.join('.')
-          else
-            path = path_parts[0]
-          end
+          path = search_param_definition['expression']
+          path = path.gsub(/.where\((.*)/, '')
+          as_type = path.scan(/.as\((.*?)\)/).flatten.first
+          path = path.gsub(/.as\((.*?)\)/, capitalize_first_letter(as_type)) if as_type.present?
```
Jammjammjamm:
```suggestion
          path = path.gsub(/.as\((.*?)\)/, as_type.capitalize) if as_type.present?
```

generator/uscore/metadata_extractor.rb
czh-orz: Can't do this because it makes the rest of string lowercase. dateTime becomes Datetime.

arscan:
Looks like this needs to be rebased?

Also, I wonder if we should have some kind of assertion that the problems do actually exist before fixing them within that method.  I just wonder what happens if/when this capability statement gets updated and we have code in there modifying the structure.  Also in the case if we try to point this at another capability statement.

czh-orz:
Yeah you have a point, but since 3.1.0 seems to be the last major update and we've stopped supporting older versions, then I thought this would be okay. Alternatively, I could go in and change the json resource files themselves.

czh-orz:
Rebased and resolved codeclimate issue.

generator/uscore/metadata_extractor.rb
```diff
@@ -320,6 +336,10 @@ def set_first_search(sequence, params)
         sequence[:searches].delete(search)
         sequence[:searches].unshift(search)
       end
+
+      def capitalize_first_letter(str)
```
radamson: Why is this needed over [`#capitalize`](https://ruby-doc.org/core-2.5.4/String.html#method-i-capitalize)?

lib/app/sequence_base.rb
```diff
@@ -714,7 +714,7 @@ def can_resolve_path(element, path)
         path_ary = path.split('.')
         el_as_array = Array.wrap(element)
         cur_path_part = path_ary.shift.to_sym
-        return false if el_as_array.none? { |el| el.try(cur_path_part).present? }
+        return false if el_as_array.none? { |el| el.send(cur_path_part).present? }
```
radamson: why send over try?

lib/app/sequence_base.rb
```diff
@@ -743,7 +743,7 @@ def resolve_element_from_path(element, path)
       def get_value_for_search_param(element)
         case element
         when FHIR::Period
-          element.start || element.end
+          'gt' + element.start || 'lt' + element.end
```
Jammjammjamm: You'll need an actual condition here because `'gt' + element.start` is always truthy.

PR #369, opened 2019-11-27
FI-509 Add optional group prefixes to ensure unique test ids.

arscan:
The problem:  In our test procedure view, you can string together multiple sequences executed against different contexts.  However, their TESTIDs remain the same, which is problematic because they are different tests because they have different inputs.

This addresses the issue by adding an optional 'prefix' field on groups in the module.yml files.  If a prefix exists, the UI (and soon the XLS outputter) appends it to the test.

Screen Shot 2019-11-27 at 11 59 53 AM

This is starting to get a little more complicated than I'm comfortable with -- you need to know to append the group id in different views.  Also, I didn't add a test to ensure this is used for uniqueness, as it is just an option.  But we may consider adding this.

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/models/module/test_case.rb
+++ b/lib/app/models/module/test_case.rb
@@ -4,15 +4,17 @@ module Inferno
   class Module
     class TestCase
       attr_accessor :id
+      attr_accessor :prefix
       attr_accessor :test_group
       attr_accessor :sequence
       attr_accessor :parameters
 
-      def initialize(id, test_group, sequence, parameters)
+      def initialize(id, test_group, sequence, index, parameters)
         @id = id
         @sequence = sequence
         @test_group = test_group
         @parameters = parameters
+        @prefix = "#{test_group.prefix}#{index}-" unless test_group.prefix.nil?
       end
 
       def title
--- a/lib/app/models/module/test_case.rb
+++ b/lib/app/models/module/test_case.rb
@@ -20,7 +20,8 @@ module Inferno
       end
 
       def prefix
-        return unless parameters.has_key?(:prefix)
+        return unless parameters.key?(:prefix)
+
         "#{parameters[:prefix]}-"
       end
 
--- a/lib/app/models/module/test_case.rb
+++ b/lib/app/models/module/test_case.rb
@@ -15,10 +15,11 @@ module Inferno
         @test_group = test_group
         @parameters = parameters
 
-        number_to_string = {}
-        ('A'..'ZZZ').each_with_index { |string, string_index| number_to_string[string_index + 1] = string }
-        index_as_string = number_to_string[index]
-        @prefix = "#{test_group.prefix}#{index_as_string}-" unless test_group.prefix.nil?
+        if test_group.prefix.present?
+          index_as_string = 'A'.dup
+          index.times { index_as_string.next! }
+          @prefix = "#{test_group.prefix}#{index_as_string}-"
+        end
       end
 
       def title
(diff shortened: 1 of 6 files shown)


lib/app/models/module/test_case.rb
```diff
@@ -4,15 +4,21 @@ module Inferno
...
+      def initialize(id, test_group, sequence, index, parameters)
         @id = id
         @sequence = sequence
         @test_group = test_group
         @parameters = parameters
+
+        number_to_string = {}
+        ('A'..'ZZZ').each_with_index { |string, string_index| number_to_string[string_index + 1] = string }
+        index_as_string = number_to_string[index]
+        @prefix = "#{test_group.prefix}#{index_as_string}-" unless test_group.prefix.nil?
```
Jammjammjamm: How about something like:
```
if test_group.prefix.present?
  index_as_string = 'A'
  index.times { index_as_string.next! }
  @prefix = "#{test_group.prefix}#{index_as_string}-"
end
```

arscan:
> Where should this change be visible? When I try to view the report, I get the following error...
> 

Fixed the report error!

Also, I added this to the onc nprm module so you can see it there now.  See screenshots in description.

lib/app/models/module/test_case.rb
arscan: This is cleaner, didn't know about `next`.  But I had to add a `.dup` because otherwise I'm mutating a frozen literal, and Rubocop had a recommendation I applied too.

Jammjammjamm:
I'm also concerned about the complexity around having to add the prefix in multiple places. I feel like the ideal thing would be for each `TestCase` to have its own unique id, but I don't know that the `TestCases` are always available in the places where we need this id, and worry that implementing it like that might be a big lift.

arscan:
> I'm also concerned about the complexity around having to add the prefix in multiple places. I feel like the ideal thing would be for each `TestCase` to have its own unique id, but I don't know that the `TestCases` are always available in the places where we need this id, and worry that implementing it like that might be a big lift.

Yeah what we really need is each 'test_case' to have a unique ID.  This basically accomplishes that by adding a prefix on the group and assigning a ID on each test case automagically.  Would you prefer if I move this out of group and into test_Case, and then updating module.yml so you can optionally ad an "id" to each sequence you name?

Jammjammjamm:
What I'd like to see is rather than setting `@prefix` on `TestCase`, that `@id` be set to include the prefix, or provide some other method to get the whole id, so that we don't have to remember to append the prefix in multiple places.

Also, what about appending a suffix rather than a prefix? Something like `OIDC-01a` instead of `SAC-OIDC-01`?

arscan:
I feel like logically this sets up like a tree, and it would be a little confusing to put it at the end like that.  The requirement is that every test must have a globally unique ID, and that meets the requirement, so it would be acceptable to me.

Another option would be to allow in the module.yml file to override the Sequence prefix with a testcase prefix.  So we could do OIDCSA-01 or OIDCa-01 or really ANYTHING-01.  Overriding is kinda how the rest of TestCase works, so maybe that is a better approach?

Jammjammjamm:
Overriding them in the yaml seems like a good solution to me.

arscan:
@Jammjammjamm I updated so you set the prefixes at the test-case level.  However, there are essentially now 2 prefixes, one at the sequence level that gets baked into the test id, and one at the testcase level that gets appended in the views.  I'm not thrilled with it, but i couldn't update the testid without doing a major rework of our whole system (because a test is essentially global, if we mucked with the testid in one test case it would update for all test cases).

I think this the representation is acceptable (e.g. SAC-OIDC-01).  OIDC-01 represents the sequence-test, and SAC represents the test case.  If you are ok with the ID format could you approve, as this is on the critical path?

PR #378, opened 2019-12-04
Fi 463 hl7 validator integration

okeefm:

This PR allows for the addition of validators beyond the internal `fhir_models` validator. The validator that is used by the system can be specified at server start, by changing the config options in `config.yml`:

```
resource_validator: external
external_resource_validator_url: http://localhost:8081
```

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-463

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -693,7 +693,7 @@ module Inferno
         end
         if p
           @profiles_encountered << p.url
-          resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(entry.resource, versioned_resource_class, p.url)
+          resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(resource, versioned_resource_class, p.url)
           errors = resource_validation_errors[:fatals]
           errors.concat resource_validation_errors[:errors]
           unless errors.empty?
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -13,7 +13,6 @@ require_relative 'utils/result_statuses'
 require_relative 'utils/search_validation'
 require_relative 'models/testing_instance'
 require_relative 'models/inferno_test'
-require_relative 'utils/base_validator'
 require_relative 'utils/hl7_validator'
 
 require 'bloomer'
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -690,12 +690,7 @@ module Inferno
         end
         if p
           @profiles_encountered << p.url
-          resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(resource, versioned_resource_class, p.url)
-          errors = resource_validation_errors[:errors]
-          unless errors.empty?
-            errors.map! { |e| "#{resource_type}/#{resource.id}: #{e}" }
-            @profiles_failed[p.url].concat(errors)
-          end
+          errors = validate_resource(resource_type, resource, p)
         else
           resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(entry.resource, versioned_resource_class)
           errors = resource_validation_errors[:errors]
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -502,7 +502,6 @@ module Inferno
           # This checks to see if the base resource conforms to the specification
           # It does not validate any profiles.
           resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(entry.resource, versioned_resource_class)
-          assert resource_validation_errors[:fatals].empty?, "Invalid #{entry.resource.resourceType}: #{resource_validation_errors[:fatals].join("<br/>\n")}"
           assert resource_validation_errors[:errors].empty?, "Invalid #{entry.resource.resourceType}: #{resource_validation_errors[:errors].join("<br/>\n")}"
 
           search_params.each do |key, value|
@@ -556,8 +555,7 @@ module Inferno
 
       def validate_resource(resource_type, resource, profile)
         resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(resource, resource_type, profile.url)
-        errors = resource_validation_errors[:fatals]
-        errors.concat resource_validation_errors[:errors]
+        errors = resource_validation_errors[:errors]
         @test_warnings.concat resource_validation_errors[:warnings]
 
         errors.map! { |e| "#{resource_type}/#{resource.id}: #{e}" }
@@ -589,7 +587,7 @@ module Inferno
           else
             warn { assert false, 'No profiles found for this Resource' }
             issues = Inferno::RESOURCE_VALIDATOR.validate(resource, versioned_resource_class)
-            issues[:fatals].concat issues[:errors]
+            issues[:errors]
           end
         end
         # TODO
@@ -693,16 +691,14 @@ module Inferno
         if p
           @profiles_encountered << p.url
           resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(resource, versioned_resource_class, p.url)
-          errors = resource_validation_errors[:fatals]
-          errors.concat resource_validation_errors[:errors]
+          errors = resource_validation_errors[:errors]
           unless errors.empty?
             errors.map! { |e| "#{resource_type}/#{resource.id}: #{e}" }
             @profiles_failed[p.url].concat(errors)
           end
         else
           resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(entry.resource, versioned_resource_class)
-          errors = resource_validation_errors[:fatals]
-          errors.concat resource_validation_errors[:errors]
+          errors = resource_validation_errors[:errors]
         end
         assert(errors.empty?, errors.join("<br/>\n"))
       end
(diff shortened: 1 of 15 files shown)


okeefm:
Side note: this PR probably should get squash-and-merged when it comes time...

lib/app/sequence_base.rb
```diff
@@ -583,7 +588,8 @@ def test_resources(resource_type)
             validate_resource(resource_type, resource, p)
           else
             warn { assert false, 'No profiles found for this Resource' }
-            resource.validate
+            issues = Inferno::RESOURCE_VALIDATOR.validate(resource, versioned_resource_class)
+            issues[:fatals].concat issues[:errors]
```
Jammjammjamm: This logic seems like it's duplicated a lot. Would it be feasible to add a method like `validation_errors` that does the validation and returns the fatal/error issues already concatenated?

lib/app/utils/fhir_models_validator.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Inferno
+  # FHIRModelsValidator extends BaseValidator to use the validation in fhir_models.
+  # It passes the validation off to the correct model version.
+  class FHIRModelsValidator
+    def initialize; end
```
Jammjammjamm: Is this actually needed?

test/unit/hl7_validator_test.rb
```diff
@@ -0,0 +1,42 @@
...
+require 'byebug'
+
+describe Inferno::HL7Validator do
+  before do
+    @validator_url = 'http://example.com:8080'
+    @validator = Inferno::HL7Validator.new(@validator_url)
+  end
+
+  describe 'Validating a good resource' do
+    it 'Shouldn\'t pass back any messages' do
```
Jammjammjamm: Use double quotes so you don't have to escape the single quote.

test/unit/hl7_validator_test.rb
```diff
@@ -0,0 +1,42 @@
...
+  describe 'Validating a bad resource' do
+    it 'Should pass back an error message' do
+      patient = FHIR::Patient.new(gender: 'robot')
+
+      stub_request(:post, "#{@validator_url}/validate")
+        .with(query: { profile: 'http://hl7.org/fhir/StructureDefinition/Patient' })
+        .to_return(status: 200, body: load_fixture('validator_bad_response'))
+      result = @validator.validate(patient, FHIR)
+
+      assert_equal 1, result[:errors].size
```
Jammjammjamm: Should this be `:fatals`?

lib/app/sequence_base.rb
okeefm: Yeah, I could look into doing that. Would cut down on repetitive code.

lib/app/utils/fhir_models_validator.rb
okeefm: No, it's not; it was there when the validator classes used to inherit from a parent class, but now that they don't this can be removed.

test/unit/hl7_validator_test.rb
okeefm: Yes; good catch.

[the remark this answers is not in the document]
okeefm: Yes; good catch.

lib/app/utils/hl7_validator.rb
```diff
@@ -0,0 +1,35 @@
...
+  class HL7Validator
+    @validator_url = nil
+
+    def initialize(validator_url)
+      @validator_url = validator_url
+    end
+
+    def validate(resource, fhir_models_klass, profile = nil)
+      profile ||= fhir_models_klass::Definitions.resource_definition(resource.resourceType).url
+      result = RestClient.post "#{@validator_url}/validate", resource.to_json, params: { profile: profile }
```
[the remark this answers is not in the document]
okeefm: Probably it would make the most sense to check in `initialize`, but that is definitely something I can do.

lib/app/utils/hl7_validator.rb
```diff
@@ -0,0 +1,35 @@
...
+      outcome = fhir_models_klass.from_contents(result.body)
+      fatals = issues_by_severity(outcome.issue, 'fatal')
+      errors = issues_by_severity(outcome.issue, 'error')
+      warnings = issues_by_severity(outcome.issue, 'warning')
+      informations = issues_by_severity(outcome.issue, 'information')
```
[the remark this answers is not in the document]
okeefm: That probably makes sense. I'll switch it.

Jammjammjamm:
Could you walk me through using the external validator? I haven't been able to get it to work.

lib/app/utils/fhir_models_validator.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def validate(resource, _fhir_version, profile_url = nil)
```
Jammjammjamm: This method signature doesn't match the one in the hl7 validator:
```
def validate(resource, fhir_models_klass, profile = nil)
```

lib/app/utils/hl7_validator.rb
```diff
@@ -0,0 +1,35 @@
...
+    @validator_url = nil
+
+    def initialize(validator_url)
+      @validator_url = validator_url
+    end
+
+    def validate(resource, fhir_models_klass, profile = nil)
+      profile ||= fhir_models_klass::Definitions.resource_definition(resource.resourceType).url
+      result = RestClient.post "#{@validator_url}/validate", resource.to_json, params: { profile: profile }
+      outcome = fhir_models_klass.from_contents(result.body)
```
Jammjammjamm: `fhir_models_klass` is coming through as a string, at least in some cases:
Screen Shot 2019-12-16 at 8 39 30 AM
I think `from_contents` is usually used on the base FHIR class, like `FHIR.from_contents`, as a way to instantiate a FHIR model without having to specify what resource type it is in advance.

okeefm: That was, in fact, the intention of that variable. Seems like sometimes that may not be what's being passed to it though.

lib/app/utils/fhir_models_validator.rb
okeefm: Good catch. Will fix.

lib/app/sequence_base.rb
okeefm: Update: after talking with Reece, I decided to condense the number of returned messages from the validators down to just `errors` and `warnings`. In the HL7 Validator, both `fatals` and `errors` are treated as `errors`, and `warnings` and `information` are treated as `warnings`. This more closely matches how we treat the messages in Inferno currently, while leaving open the ability to have more granular error reporting in the future.

lib/app/utils/hl7_validator.rb
okeefm: @Jammjammjamm If you get a chance, can you verify that you can no longer reproduce this issue?

Jammjammjamm: The issue has changed:
Screen Shot 2019-12-18 at 12 22 23 PM
It should be `FHIR.from_contents` or `FHIR::DSTU2.from_contents`, not `FHIR::SomeResourceType.from_contents`. Or if you want to instantiate that specific class you should be able to use something like`klass.new(JSON.parse(whatever))`.

okeefm: This issue should be fixed. The only note I have is that the validation will still be a touch off with DSTU2 resources (though it won't error out), because the FHIR Validator service that we've built only supports R4 at the moment.

lib/app/sequence_base.rb
```diff
@@ -686,13 +690,15 @@ def check_resource_against_profile(resource, resource_type, specified_profile =
         end
         if p
           @profiles_encountered << p.url
-          errors = p.validate_resource(resource)
+          resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(resource, versioned_resource_class, p.url)
```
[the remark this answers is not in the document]
okeefm: It looks like `validate_resource` has different side effects than this bit of code,  in that `@test_warnings` gets the warnings in `validate_resource`. As long as that's ok to do in this context, I don't see why not.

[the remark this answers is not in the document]
okeefm: Paging @Jammjammjamm and/or @radamson for their opinion on whether we're ok to use `validate_resource` here.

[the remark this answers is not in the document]
arscan: This whole method (`check_resource_against_profile`) is very suspect.  It is only used in the DSTU2 argo medication sequences, and it has a check at the beginning that requires the resources to be DSTU2 (so this has a very specific use and isn't particularly generic).  I believe that @okeefm was just swapping in the new API for the old one, so he changed the one line to keep the existing behavior (with the new API), right?

`sequence_base` has way too many methods like this that I would like to see cleaned up.  But I'm not sure that's in the scope of this ticket.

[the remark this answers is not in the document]
Jammjammjamm: Using `validate_resource` in this branch of the condition looks like it should be fine to me. It would be nice if `validate_resource` didn't require a profile and we could use it in both branches of the condition, but I don't think that needs to be done in this PR.

PR #382, opened 2019-12-06
Fi 530 revinclude skip

arscan:
If revinclude doesn't return anything, we should skip as it just means they have an incomplete record.  At the very least we should skip if there are no resources of the type you are revincluding against (e.g. if there are no AllergyIntolerances, there should not be any provenances either), but this update should handle both cases.

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -220,7 +220,7 @@ module Inferno
                 assert_response_ok(reply)
                 assert_bundle_response(reply)
                 #{resource_variable} = reply&.resource&.entry&.map(&:resource)&.any? { |resource| resource.resourceType == '#{resource_name}' }
-                assert #{resource_variable}, 'No #{resource_name} resources were returned from this search'
+                skip 'No #{resource_name} resources were returned from this search' unless #{resource_variable}
           )
         end
         sequence[:tests] << revinclude_test
--- a/lib/modules/uscore_v3.1.0/bodyheight_sequence.rb
+++ b/lib/modules/uscore_v3.1.0/bodyheight_sequence.rb
@@ -307,7 +307,7 @@ module Inferno
         assert_response_ok(reply)
         assert_bundle_response(reply)
         provenance_results = reply&.resource&.entry&.map(&:resource)&.any? { |resource| resource.resourceType == 'Provenance' }
-        assert provenance_results, 'No Provenance resources were returned from this search'
+        skip 'No Provenance resources were returned from this search' unless provenance_results
       end
 
       test 'Observation resources returned conform to US Core R4 profiles' do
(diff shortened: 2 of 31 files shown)


generator/uscore/uscore_generator.rb
```diff
@@ -217,7 +217,7 @@ def create_revinclude_test(sequence)
                 assert_response_ok(reply)
                 assert_bundle_response(reply)
                 #{resource_variable} = reply&.resource&.entry&.map(&:resource)&.any? { |resource| resource.resourceType == '#{resource_name}' }
-                assert #{resource_variable}, 'No #{resource_name} resources were returned from this search'
+                skip 'No #{resource_name} resources were returned from this search' unless #{resource_variable}
```
Jammjammjamm: I think this was correct as it was. This change skips the test if no Provenance resources are returned, not if none of the resources the Provenance resources should refer to are found. The search and interaction tests skip using this when no resources have been found:
```
skip 'No resources appear to be available for this patient. Please use patients with more information.' unless @resources_found
```

czh-orz:
It already skips if no resource was found during this line:
`search_params.each { |param, value| skip "Could not resolve #{param} in given resource" if value.nil? }`.

generator/uscore/uscore_generator.rb
arscan: revInclude is weird and a little mind-bendy, but in practice for US Core, using AllergyIntolerance (as an example) we should:
1) skip this test if no AllergyIntolerances exist for the patient.  I agree that skip line you provided would accomplish this, but I'm not seeing that line in the test.  And my update doesn't directly test this, but instead assumes that if you don't have any provenance, then you won't have any AlergyIntolerance, and then it skips.
2) skip this test if no Provenances are linked to any AllergyIntollerances for this patient.  We need to know they support provenances for AllergyIntolerances, so skip if we don't see any.

I could be missing something big, obviously.

arscan:
> It already skips if no resource was found during this line:
> `search_params.each { |param, value| skip "Could not resolve #{param} in given resource" if value.nil? }`.

where is that?  I'm looking at this generated file: https://github.com/onc-healthit/inferno/blob/development/lib/modules/uscore_v3.1.0/us_core_allergyintolerance_sequence.rb#L173

czh-orz:
You can see it sequences where the first search isn't a search by just patient. For example, CarePlan or SmokingStatus.

generator/uscore/uscore_generator.rb
Jammjammjamm: Yeah, I think we need two skips in order for the skips to be clear and not misleading:

- One if no resources of this type have been found at all (could use the line I quoted)
- One if no Provenance resources are found.

Jammjammjamm: Yeah, I think we need two skips in order for the skips to be clear and not misleading:

- One if no resources of this type have been found at all (could use the line I quoted)
- One if no Provenance resources are found.

arscan:
We have another PR that works in this space #389 and we should just roll this into that update, and closs this.

arscan:
I'm pulling this one in, and then the other PR addresses some of the other comments.

PR #384, opened 2019-12-09
Fi 370 vital signs sequences

czh-orz:
This PR creates sequences for the vital signs profiles that us core supports. 

**Submitter:**
- [] Tests are included and test edge cases

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


Jammjammjamm:
This needs to be rebased and the generator rerun so that the search unit tests are included.

lib/app/utils/validation.rb
```diff
@@ -61,7 +61,17 @@ def self.get_resource(json, version)
       lab_results: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-lab',
       pediatric_bmi_age: 'http://hl7.org/fhir/us/core/StructureDefinition/pediatric-bmi-for-age',
       pediatric_weight_height: 'http://hl7.org/fhir/us/core/StructureDefinition/pediatric-weight-for-height',
-      pulse_oximetry: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-pulse-oximetry'
+      pulse_oximetry: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-pulse-oximetry',
+      vitals_panel: 'http://hl7.org/fhir/StructureDefinition/vitalspanel',
```
Jammjammjamm: The vitals panel, oxygen saturation, and BMI profiles are not mentioned in the list of vital signs that are part of USCDI (http://hl7.org/fhir/us/core/general-guidance.html#us-core-data-for-interoperability-and-2015-edition-common-clinical-data-set).

lib/app/utils/validation.rb
```diff
@@ -163,5 +181,9 @@ def self.guess_r4_profile(resource)
 
       candidates.first
     end
+
+    def self.resource_contains_code(resource, code)
```
Jammjammjamm: This method doesn't seem as general as its name implies. Maybe `observation_contains_code` until it becomes more general?

lib/app/utils/validation.rb
arscan: It seems reasonable to remove the ones that are not listed in that table for now.  We can turn this back on easily later if we want by updating the capability statement for the server.

lib/app/utils/validation.rb
arscan: agreed, fixed.

lib/app/utils/validation.rb
arscan: Note: we just have it not generate the sequences.  We left them in the profiles and validation because it doesn't seem like it hurts anything in case we need to re-add (though it might be cruft in the future).

PR #385, opened 2019-12-10
Allow Inferno to be used as a library

radamson:
This PR is the first step in allowing Inferno to be used as a library in other applications.  It mostly just moves files around and adds a `.gemspec`.  There is a lot more that could be changed to create an API that really empowers developers to create their own modules, but those are expected to come in future PRs that build off this.

-  Moves all the files in `lib` to `lib/inferno` so that Inferno can be used as a library in other applications with `require inferno`
- Adds `inferno.gemspec`
- Adds class variable and accessors `@@inferno_modules` to `Endpoint` so modules can be registered and appear on the landing page.

```ruby
require 'example/version'
require 'inferno'
Dir.glob(File.join(__dir__, 'modules', '**', '*_sequence.rb')).each { |file| require file }

module Example

  def self.start_app

    # Load the module
    Dir.glob(File.join(__dir__, 'modules', '*_module.yml')).each do |file|
      this_module = YAML.load_file(file).deep_symbolize_keys
      Inferno::Module.new(this_module)
    end

    # Add to module list as the only module (can << if you want to keep Inferno defaults)
    example_module = Inferno::Module.get('example_read')
    Inferno::App::Endpoint.inferno_modules = [ example_module ]
    Inferno::App.new
  end
end
```

This separate app has its own module, sequences and `config.ru` file for running the app.

I can put a full example up illustrating this if desired.

**Submitter:**
- [ ] This pull request describes why these changes were made
- [ ] Internal ticket for this PR:

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


arscan:
In gems are you allowed to include files that aren't under the 'lib/inferno' directory?  My understanding is that Ruby's library system works by essentially adding 'lib' to Ruby's internal path, so when you `require 'inferno'` it essentially just looking for a "inferno.rb" within the running path, which (yay) there is.  Typically then that calls all the files within `/lib/inferno` to keep everything separate.  I have no idea what happens if you start referencing files that aren't within that directory -- could they collide with files from other gems?

Also, I'm not sure we can call this 'inferno' because in RubyGems there is already an inferno.

And finally, a common pattern is to export things as a cli tool, so you can encapsolate all that boilerplate into a command line call.  Like `bundle exec inferno start` launches the web server that looks for all module files in the current directory, and auto-imports everything under `/sequences` or whatever.  No "rackup" needed (what layperson would even know what rackup is?)

Broadly I think we *need* to do this, I just get stuck on some specifics (like my questions above) which causes this to be a much bigger lift all at once.

radamson:
My understanding is adding inferno to the [`Gemfile` adds it to the load path](https://bundler.io/v1.5/bundler_setup.html#setting-up-your-application-to-use-bundler), which then let's us use require on it.  `inferno.rb` requires `app` which then loads everything else (it was our previous starting point so just requiring that caused the least amount of changes).

I'm not exactly sure why you would want to reference other files that are not in that directory.  I was still able to do something like `require inferno/version` if I only wanted to load that file, but you are be able to use `Inferno::Version` just by requiring `inferno` as that includes `version` (plus all the other classes like SequenceBase that are needed).  All the Inferno capabilities are namespaced under the `Inferno` namespace so there shouldn't be any collisions unless there is another `Inferno` gem being used.

Agreed it looks like it needs to be renamed.  I'm happy to rename it, but this at least let me use it in another repo by locally linking to it in the Gemfile.  We could rename to `inferno-community` and and create that repo in the `inferno-community` org?

I could add the CLI, but I think that would be better for a future PR.  This one really is only to reorg and add the gemspec so it can be included as a library.  Getting this in now lets us stay agile so at least let's folks can get started while not preclude adding the executable stuff later.

arscan:
> I'm not exactly sure why you would want to reference other files that are not in that directory. You still can do something like `require inferno/version` if you only wanted to load that file, but you would typically use `Inferno::Version` or any of the other classes like `SequenceBase` just by requiring `inferno`. All the Inferno capabilities are namespaced under the `Inferno` namespace so there shouldn't be any collisions unless there is another `Inferno` gem being used.

I was just wondering if it was a problem that not all the files that are referenced internally in the gem (e.g. resource files) live under `lib/inferno`, and if that could cause collisions with other gems or something.  I guess as long as it gets added to the gem somehow (in your method, adding everything that is checked into git) its fine.

radamson:
Ah sorry, I see what you mean now about the file references.  I don't think it would cause collisions based on looking at other gems in my installation directory, but now I'm worried those might be lost when we actually create a real gem compared to just the relative path.  In my example app those files, like `public` which has all the jquery to power to UI, seemed to be working without me having to do anything special.

I'll try building and installing the gem locally to see what happens.

radamson:
I created a gem with `gem build inferno.gemspec`, moved it over to the example app and installed it with `gem install ./inferno-2.8.0.gem` and `bundle install --local`.  The app still worked fine.

Looking at the `$LOAD_PATH` I see the inferno gem listed:

`.../ruby/gems/2.5.0/gems/inferno-2.8.0/lib`

With only the lib directory included I don't think there will be any collisions.  I'm happy moving the files into the `lib/inferno` if desired though.

arscan:
I know ruby uses `_` as a word delimiter, but please use `-` for repo names because that's what we've standardized on.

So in `.../ruby/gems/2.5.0/gems/inferno-2.8.0/` do you see all the other directories/files (`public`, etc)?  If so, then I think we are good on this issue. I do think it might go against convention a bit, but if it works then I don't have a problem with doing it this way as a first step.

This is a big change and would require coordination and agreement from the whole team.

radamson:
Renamed the repo 👍.

Yes all the other directories and files are in there.  It grabs everything checked into git, similar to how we handle `fhir_client`.  I included the `inferno_core-2.8.0.gem` in `inferno-starter` which you can open up to see what it looks like.  I agree having some of these files outside the `lib` seems odd, but I'm ok with whatever works best for the team.

.rubocop_todo.yml
```diff
@@ -77,9 +143,9 @@ Style/RegexpLiteral:
   Exclude:
     - 'test/test_helper.rb'
 
-# Offense count: 542
+# Offense count: 5994
```
Jammjammjamm: Something seems wrong here.

lib/inferno/app/endpoint.rb
```diff
@@ -1,16 +1,20 @@
...
+require_relative 'models/module'
+
 module Inferno
   class App
     class Endpoint < Sinatra::Base
+
       register Sinatra::ConfigFile
 
-      config_file File.join('..', '..', ENV['INFERNO_CONFIG_FILE'] || 'config.yml')
+      config_file File.join('..', '..', '..', ENV['INFERNO_CONFIG_FILE'] || 'config.yml')
```
Jammjammjamm: I think we'll need a new methodology for this if Inferno is used as a gem.

lib/inferno/app/endpoint.rb
```diff
@@ -53,23 +57,32 @@ class Endpoint < Sinatra::Base
...
+        @@inferno_modules
+      end
+
+      def self.inferno_modules=(new_modules)
+        @@inferno_modules = new_modules
+      end
+
       def render_index
-        unless defined?(settings.presets).nil? || settings.presets.nil?
+        unless @@inferno_modules
```
Jammjammjamm: What does this change do?

.rubocop_todo.yml
radamson: Huh.  Not sure why that's happening as it is mostly just moving files... I'll investigate.

lib/inferno/app/endpoint.rb
radamson: Agreed.  I can work on that if it belongs in this PR.  I was trying to minimize the amount of changes as much as I could because a lot more things (like `requires`, the erb templates, and the testing instance models) also have to change, but it would be a huge overhaul at once.

I'll look into improving this for now though!

lib/inferno/app/endpoint.rb
radamson: There was no way to change the modules available in the UI except through the config file.  This change, and few methods above it, allowed the modules to be changed programmatically with

```ruby
    # Add to module list as the only module (can << if you want to keep Inferno defaults)
    example_module = Inferno::Module.get('example_read')
    Inferno::App::Endpoint.inferno_modules = [ example_module ]
    Inferno::App.new
```

```ruby
Inferno::App::Endpoint.inferno_modules << example_module
``` 
also works if you want to append on what is already there (i.e. what is in the default inferno config)

PR #387, opened 2019-12-13
fi-538 Add generator parameters for prefix and optional tests

czh-orz:
This PR updates the generator rake task to take in two more parameters. The first is prefix, which will prepend all sequence and module names with the given prefix string. For example, setting a prefix of 'abc' will generate abc_sequence_name.rb and abc_module_name.yml. 

The second is to disable optional tests. Set this to 'true' if you want to not generate optional tests within the sequences. 

`bundle exec rake inferno:generate['uscore','uscore_v3.1.0', 'prefix_here', 'disable_optional_tests']`

This will default to the original behavior if none of the two new parameters are set. So if you use 
`bundle exec rake inferno:generate['uscore','uscore_v3.1.0']`, the same old sequences will be generated. 

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-538

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


czh-orz:
rebased and added prefix parameter

generator/generator_base.rb
```diff
@@ -76,7 +80,7 @@ def module_yml_out_path
       end
 
       def sequence_out_path
-        File.expand_path("#{module_yml_out_path}/#{path}")
+        File.expand_path("#{module_yml_out_path}/#{prefix + '_' if prefix.present?}#{path}")
```
Jammjammjamm: I think it would be nice to have a method like this so that this logic doesn't have to be repeated:
```
def prefix_string
  prefix.present? ? prefix + '_' : ''
end
```

czh-orz: This is fixed. I also changed it to behave like before when none of the new params are specified.

Jammjammjamm: Could you update the PR description so that its instructions match the current implementation? If I want to remember how to use this functionality I'm going to find this PR and look at those instructions.

Jammjammjamm:
The module yaml file needs to use the prefix also. I generated the US Core sequence with a prefix of `abc`:
```
ᐅ git status
Changes not staged for commit:

	modified:   lib/modules/uscore_v3.1.0_module.yml

Untracked files:

	lib/modules/abc_uscore_v3.1.0/
```

czh-orz:
Added the prefix to the module file and the unit tests.

Jammjammjamm:
I think this is good once the remaining codeclimate issue is fixed.

Jammjammjamm:
Superceded by #398

PR #395, opened 2019-12-17
FI-507: Automatic token refresh

Jammjammjamm:
Inferno's tests can take some time, so Inferno needs to be able to automatically refresh access tokens before they expire.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-507

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/modules/smart/shared_launch_tests.rb
+++ b/lib/modules/smart/shared_launch_tests.rb
@@ -55,7 +55,11 @@ module Inferno
         @instance.update(encounter_id: token_response_body['encounter']) if token_response_body['encounter'].present?
 
         required_keys = ['token_type', 'scope']
-        required_keys << 'expires_in' if require_expires_in
+        if require_expires_in
+          required_keys << 'expires_in'
+        else
+          warning { assert expires_in.present?, 'Token exchange response did not contain the recommended `expires_in` field' }
+        end
 
         required_keys.each do |key|
           assert token_response_body[key].present?, "Token response did not contain #{key} as required"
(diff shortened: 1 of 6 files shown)


lib/modules/smart/shared_launch_tests.rb
```diff
@@ -40,7 +40,16 @@ def validate_token_response_contents(token_response, require_expires_in:)
 
         assert token_response_body['access_token'].present?, 'Token response did not contain access_token as required'
 
-        @instance.update(token: token_response_body['access_token'], token_retrieved_at: DateTime.now)
+        expires_in = token_response_body['expires_in']
+        if expires_in.present? # rubocop:disable Style/IfUnlessModifier
+          assert expires_in.is_a?(Numeric), "`expires_in` field is not a number: #{expires_in.inspect}"
```
arscan: The requirement behind here is weaker than I'd like:

https://tools.ietf.org/html/rfc6749#section-5.1

"Numerical values are included as JSON numbers."

But the definition of expires_in is never explicitely called numeric.

"RECOMMENDED.  The lifetime in seconds of the access token.  For
         example, the value "3600" denotes that the access token will
         expire in one hour from the time the response was generated.
         If omitted, the authorization server SHOULD provide the
         expiration time via other means or document the default value."

arscan: More importantly, this assert needs to move to the end of the test somehow, because now we aren't saving patient_id and the system is blocked from testing any further until this is fixed (because no patient_id)

arscan: I think a reasonable fix is to turn this into a warning (because the requirement is slightly vague), which allows the later code to run and the issue is noted.

Jammjammjamm: Yeah, this assertion was really here because `token_expires_in` is an integer field and updating it with a non-integer value raises an error. I'll add some logic to prevent that and change this to a warning.

PR #400, opened 2019-12-24
Fi 481 create docref test

czh-orz:
This PR adds a test for the docref operation. It also adds the ability to check the server capabilities for resource operation support. The test itself skips if docref is not supported in the capability statement. It performs the GET request and asserts it's 200 response. I don't check if anything is returned because returning an empty bundle is valid when the server can't generate the CCD. 

**Submitter:**
- [x ] Internal ticket is properly labeled (Community/Program)

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -72,6 +72,12 @@ module Inferno
               create_interaction_test(sequence, interaction)
             end
 
+          sequence[:operations]
+            .select { |operation| operation[:expectation] == 'SHALL' }
+            .each do |operation|
+              create_docref_test(sequence) if operation[:operation] == 'docref'
+            end
+
           create_include_test(sequence) if sequence[:include_params].any?
           create_revinclude_test(sequence) if sequence[:revincludes].any?
           create_resource_profile_test(sequence)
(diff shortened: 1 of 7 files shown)


generator/uscore/uscore_generator.rb
```diff
@@ -175,6 +181,24 @@ def create_authorization_test(sequence)
...
+        docref_test = {
+          tests_that: 'The server is capable of returning a reference to a generated CDA document in response to the $docref operation',
+          index: sequence[:tests].length + 1,
+          link: 'http://hl7.org/fhir/us/core/2019Sep/CapabilityStatement-us-core-server.html#documentreference',
+          description: 'A server SHALL be capable of responding to a $docref operation and capable of returning at least a reference to a generated CCD document, if available.'
+        }
+
+        docref_test[:test_code] = %(
+          skip_if_not_supported(:#{sequence[:resource]}, [], [:docref])
+          search_string = "/DocumentRefernce/$docref?patient=\#{@instance.patient_id}"
```
arscan: Typo on Resource name!

czh-orz: ........ fixed!

arscan:
I think we can add a bit more meat to this test based on the spec: https://www.hl7.org/fhir/us/core/OperationDefinition-docref.html

Ideas to make this a bit more thorough:
* Is both POST and GET approaches required? (hard to tell in a quick scan) if so easy enough to check both ways
* check to make sure a bundle is returned
* we should skip if the bundle returned is empty
* check that everything in the bundle validates to the documentreference profile (or add them to a list to validate that later? i forget how we approach this)
* we should follow the reference if possible to make sure that it actually returns a file?  Probably attach the bearer token to that request?

PR #416, opened 2020-01-14
Fi 540 check resource id from read test

czh-orz:
This PR makes it so that validate_read_reply will check if the resource returned has the same id as the resource used for the read test. I don't test the case when the resource is a reference because I didn't know a way to get the id for any reference and I don't think that situation is used in any of our existing sequences.

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-540

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/app/sequence_base.rb
```diff
@@ -524,7 +524,8 @@ def validate_search_reply(klass, reply, search_params)
       def validate_read_reply(resource, klass)
         class_name = klass.name.demodulize
         assert !resource.nil?, "No #{class_name} resources available from search."
-        if resource.is_a? FHIR::DSTU2::Reference
+        resource_is_reference = resource.is_a? FHIR::DSTU2::Reference
```
arscan: Does this line only work for DSTU2 servers?  The name of the method doesn't seem to indicate it is only for DSTU2.

arscan: Could you use `resource.is_a? versioned_resource_class('Reference')`

lib/app/sequence_base.rb
```diff
@@ -524,7 +524,8 @@ def validate_search_reply(klass, reply, search_params)
...
+        if resource_is_reference
           read_response = resource.read
         else
           id = resource.try(:id)
```
arscan: Might as well change this `resource&.id` maybe

Jammjammjamm:
I'm not saying that you need to support getting ids from references, but if you want to, the id is whatever comes after the last `/` in the reference. If we don't ever use references here, though, it would be nice to get rid of that code altogether.

lib/app/sequence_base.rb
czh-orz: Changed.

lib/app/sequence_base.rb
czh-orz: Removed logic for reference altogether since it's not being used anywhere.

czh-orz:
> I'm not saying that you need to support getting ids from references, but if you want to, the id is whatever comes after the last `/` in the reference. If we don't ever use references here, though, it would be nice to get rid of that code altogether.

I just removed the reference part.

czh-orz:
I added the reference check back in because it's being used in one test in the Argonaut MedicationOrder sequence.

PR #419, opened 2020-01-21
Fi 601 Pass comma-separated patients to US Core

arscan:
Pass a comma separated list of patients to US Core.  A few loose ends need fixing, and unit tests need to be updated.  For discussion.  FYI @czh-orz @Jammjammjamm 

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


generator/uscore/uscore_generator.rb
```diff
@@ -250,10 +258,17 @@ def create_revinclude_test(sequence)
...
                 #{resource_variable} = fetch_all_bundled_resources(reply.resource).select { |resource| resource.resourceType == '#{resource_name}'}
-                skip 'No #{resource_name} resources were returned from this search' unless #{resource_variable}.present?
+                any_provenances ||= provenance_results.present?
                 #{resource_variable}.each { |reference| @instance.save_resource_reference('#{resource_name}', reference.id) }
           )
         end
+
+        revinclude_test[:test_code] += %(
+          end
+          skip 'No Provenance resources were returned from this search' unless any_provenances
```
Jammjammjamm: It would be nice to use our `skip_unless` helper wherever possible.

generator/uscore/uscore_generator.rb
```diff
@@ -240,7 +242,13 @@ def create_revinclude_test(sequence)
           description: "A Server SHALL be capable of supporting the following _revincludes: #{sequence[:revincludes].join(', ')}"
         }
         search_params = get_search_params(first_search[:names], sequence)
-        revinclude_test[:test_code] = search_params
+        revinclude_test[:test_code] = %(
+               any_provenances = false
+               could_not_resolve_all = []
+               resolved_one = false
```
Jammjammjamm: These aren't always needed, for example in the AllergyIntolerance sequence.

generator/uscore/uscore_generator.rb
```diff
@@ -250,10 +258,17 @@ def create_revinclude_test(sequence)
                 assert_response_ok(reply)
                 assert_bundle_response(reply)
                 #{resource_variable} = fetch_all_bundled_resources(reply.resource).select { |resource| resource.resourceType == '#{resource_name}'}
-                skip 'No #{resource_name} resources were returned from this search' unless #{resource_variable}.present?
+                any_provenances ||= provenance_results.present?
```
Jammjammjamm: Should we `break` out of the iteration once we've found some?

generator/uscore/uscore_generator.rb
```diff
@@ -285,20 +300,27 @@ def create_search_test(sequence, search_param)
 
         is_first_search = search_param == find_first_search(sequence)
 
+        comparator_search_code = get_comparator_searches(search_param[:names], sequence)
+
         search_test[:test_code] =
           if is_first_search
+            # rcs question: are comparators ever be in the first search?
```
Jammjammjamm: They are not currently and I'd say they are unlikely to be in the future.

generator/uscore/uscore_generator.rb
```diff
@@ -740,15 +786,26 @@ def get_search_params(search_parameters, sequence, grab_first_value = false)
 
         if search_param_string.include? 'get_value_for_search_param'
           search_param_value_check =
-            %(search_params.each { |param, value| skip "Could not resolve \#{param} in given resource" if value.nil? }\n)
-          return [search_param_string, search_param_value_check].join('')
+            %(
+              if search_params.any? { |param, value| value.nil? }
```
Jammjammjamm:
```suggestion
              if search_params.any? { |_param, value| value.nil? }
```
This could also be `if search_params.value? nil`, but I think that is less clear.

generator/uscore/uscore_generator.rb
```diff
@@ -703,31 +742,38 @@ def get_first_search_with_fixed_values(sequence, search_parameters, save_resourc
...
-            break#{' if values_found == 2' if find_two_values}
+              save_resource_ids_in_bundle(#{save_resource_ids_in_bundle_arguments})
+              save_delayed_sequence_references(@#{sequence[:resource].underscore}_ary[patient])
+              validate_search_reply(versioned_resource_class('#{sequence[:resource]}'), reply, search_params)
+              #{'test_medication_inclusion(@medication_request_ary[patient], search_params)' if sequence[:resource] == 'MedicationRequest'}
+              break#{' if values_found == 2' if find_two_values}
+            end
           end
           #{skip_if_not_found(sequence)})
+          #{skip_if_could_not_resolve}
```
Jammjammjamm: Is this intentionally a comment rather than part of the string?

generator/uscore/uscore_generator.rb
```diff
@@ -240,7 +242,13 @@ def create_revinclude_test(sequence)
           description: "A Server SHALL be capable of supporting the following _revincludes: #{sequence[:revincludes].join(', ')}"
         }
         search_params = get_search_params(first_search[:names], sequence)
-        revinclude_test[:test_code] = search_params
+        revinclude_test[:test_code] = %(
+               any_provenances = false
+               could_not_resolve_all = []
```
Jammjammjamm: Could this be given a more clear name? Maybe something along the lines of `required_search_params`?

generator/uscore/uscore_generator.rb
```diff
@@ -659,22 +688,32 @@ def fixed_value_search?(search_parameters, sequence)
 
       def get_first_search_by_patient(sequence, search_parameters, save_resource_ids_in_bundle_arguments)
         %(
-          #{get_search_params(search_parameters, sequence)}
-          reply = get_resource_by_params(versioned_resource_class('#{sequence[:resource]}'), search_params)
-          assert_response_ok(reply)
-          assert_bundle_response(reply)
+          @#{sequence[:resource].underscore}_ary = {}
```
Jammjammjamm: Since this is now a hash, I'd love it if it wasn't named `*_ary`. The US Core resources work out so that `*s` works as a plural for them all.

lib/modules/uscore_v3.1.0/test/fixtures/bmi.json
```diff
@@ -1,46 +0,0 @@
-{
```
Jammjammjamm: These fixtures are not generated and need to be restored.

generator/uscore/templates/sequence.rb.erb
```diff
@@ -9,14 +9,18 @@ module Inferno
...
-      requires :token<%=", :patient_id" unless delayed_sequence%>
+      requires :token<%=", :patient_ids" unless delayed_sequence%>
       conformance_supports :<%=resource%><%="
       delayed_sequence" if delayed_sequence%>
 <%=search_validator%>
       details %(
         The #{title} Sequence tests `#{title.gsub(/\s+/, '')}` resources associated with the provided patient.
       )
 
+      def patient_ids
```
Jammjammjamm: Should this go in SequenceBase? I know putting more stuff in SequnceBase is unfortunate, but I think it makes sense since this method seems necessary to support the new field in TestingInstance.

generator/uscore/uscore_generator.rb
```diff
@@ -759,7 +816,7 @@ def get_search_param_hash(search_parameters, sequence, grab_first_value = false)
         search_parameters.each_with_object({}) do |param, params|
           params[param] =
             if param == 'patient'
-              '@instance.patient_id'
+              'patient'
```
Jammjammjamm: What's this change?

arscan: Hah, we should just remove it.  I didn't look too closely at the logic, was doing a replace any time we are doing `@instance.patient_id` because they are now in a loop block where `patient` is defined.

generator/uscore/uscore_generator.rb
arscan: I was wondering if this would make more sense, and quickly scanned to see if we could just replace with 's'.  But I didn't end up doing it because it is easy to drop an `s` in places by mistake in the generator.  But I can readd.

generator/uscore/uscore_generator.rb
```diff
@@ -703,31 +742,38 @@ def get_first_search_with_fixed_values(sequence, search_parameters, save_resourc
...
+          @#{sequence[:resource].underscore}_ary = {} 
+          @resources_found = false
           #{'values_found = 0' if find_two_values}
           #{values_variable_name} = [#{search_param[:values].map { |val| "'#{val}'" }.join(', ')}]
-          #{values_variable_name}.each do |val|
-            search_params = { 'patient': @instance.patient_id, '#{search_param[:name]}': val }
-            reply = get_resource_by_params(versioned_resource_class('#{sequence[:resource]}'), search_params)
-            assert_response_ok(reply)
-            assert_bundle_response(reply)
+          could_not_resolve_all = []
```
czh-orz: We don't need the resolved_one logic in any of the first searches because we're not resolving the values from the resources. They're all either just patient or patient + some fixed value.

czh-orz:
I think the only thing that needs to loop through patients are the first searches for patient-based resources. Once all the resources are in the array, we can use `resolve_element_from_path` to look through all the resources. Then pretty much nothing else has to change I think.

generator/uscore/templates/sequence.rb.erb
arscan: I don't think so -- to me it is a matter of scope. Should all tests everywhere need access to this method?  I don't think so.  Yeah it's unfortunate that all testing instances will have this field, but hopefully we can fix that at some point.

arscan:
Note: temporarily removed the generated unit tests because major refactor needed.  We have decided to go this route though so our plan is to merge to allow end-to-end tests with other downstream updates and backfill unit tests to catch edge cases over the coming weeks.

PR #433, opened 2020-02-03
FI-590: Add input for Device type

Jammjammjamm:
This branch adds an input field for Device type. This is needed because the US Core Implantable Device profile only applies to implantable devices, but without specifying a Device type, there is no way for Inferno to differentiate between implantable and non-implantable devices.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-590

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/templates/sequence.rb.erb
+++ b/generator/uscore/templates/sequence.rb.erb
@@ -13,7 +13,7 @@ module Inferno
 
       test_id_prefix '<%=test_id_prefix%>'
 
-      requires :token<%=", :patient_ids" unless delayed_sequence%>
+      requires :token<%=", :patient_ids" unless delayed_sequence%><%=', :device_system, :device_code' if resource == 'Device'  %>
       conformance_supports :<%=resource%><%="
       delayed_sequence" if delayed_sequence%>
 <%=search_validator%>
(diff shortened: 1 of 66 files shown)


arscan:
To keep the UI simpler, can we just ask for the code and not bother with the system?  We are really just looking for a way to filter down the results to known implantable device types, and it seems unlikely that they would use the same code repeated in two separate codesystems in a small handful of patient records to make this technically needed.

And if that sounds reasonable, I think it would be worthwhile to accept a comma separated list of codes to filter by instead of a single code, so you could paste in a long list of allowable device types and just validate any that are in that set.

arscan:
If you run it without the device code filter, then rerun it with a bad filter (so none should come back), it still seems to know about the device it saw previously.  Probably because we save references in the db and don't clear them?  I don't know exactly how we want to handle this... might be a real pain.

Jammjammjamm:
Making the Device sequence destroy all of the Device references when it starts wouldn't be difficult, just another special case to add.

arscan:
I think a special case to clear device references is fine.  Just don't do it on all the us core profile tests (unless we really analyze the impact globally).

Jammjammjamm:
@arscan The issue you raised should be addressed now.

PR #436, opened 2020-02-06
FI-584: Validate CapabilityStatement

Jammjammjamm:
This branch makes test C-02 validate the server's CapabilityStatement. Before, it just checked that a CapabilityStatement model could be populated from the server's response without actually validating the contents of the CapabilityStatement. 

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-584

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/modules/core/capability_statement_sequence.rb
```diff
@@ -127,6 +127,10 @@ def json_formats
         rescue StandardError
           assert false, 'Capability Statement could not be parsed.'
         end
+
+        issues = Inferno::RESOURCE_VALIDATOR.validate(@conformance, versioned_resource_class)
```
okeefm: Would this make more sense in the `assert_valid_conformance` method? Or is this out of scope for that?

Jammjammjamm: I initially tried that, but `assert_valid_conformance` is used in the remaining conformance tests to check that we have received a CapabilityStatement that we can parse. I don't think we want all of those tests failing if we receive an invalid CapabilityStatement. It probably makes the most sense to rename that method to something like `assert_conformance_received`.

okeefm: That makes sense to me. Up to you whether you want to rename the method; I'm good approving it as-is.

PR #441, opened 2020-03-03
Generate jwks

tstrass:
## Summary
- JWK generation for the purpose of authenticating with a SMART FHIR authorization server. The JWK is written to a file in the root of the repo called `keyfile`, which is gitignored.
  - If Inferno finds that file it will use the existing key, and if the file doesn't exist, it will generate a new key and create the file.
  - I arbitrarily chose a 2048 bit RSA key. ECDSA was also an option.

- New endpoint `/jwks` which serves the JSON Web Key Set containing Inferno's public key.

## Testing
- Run Inferno, and navigate in the browser to `/jwks`. Confirm the JWKS is served.
- Remove `keyfile`.
- Navigate to `/jwks` again and confirm a new key has been generated.
- Kill and rerun Inferno.
- Confirm the same JWKS is still served at `/jwks`

## Reference
- [SMART Backend Services Spec](http://hl7.org/fhir/uv/bulkdata/authorization/index.html#registering-a-smart-backend-service-communicating-public-keys)
- [RFC 7517 JSON Web Key (JWK)](https://tools.ietf.org/html/rfc7517)


tstrass:
My bad, meant for this PR to be into a branch in our fork. Please ignore.

PR #450, opened 2020-04-10
Fix Zulip chat link

radamson:
The Zulip link was broken in the README.  This fixes it.

I suspect that the link may have changed when HL7 merged their Zulip channels, but that's just a guess.

Addresses https://github.com/onc-healthit/inferno/issues/444.


radamson:
PR already exists to address this.

PR #470, opened 2020-07-16
Fi 900 - generic ui fields - proof of concept

czh-orz:
This is adds a way for sequences to specify their own prerequisites without changing the the view erb files or changing the testinginstance data model. They would just "require" whatever prerequisites they need and the UI and database entry for it would be made by Inferno. I have it working with US Core tests. The patient id's field and device codes field are not in the default.erb file, but you can see them in the UI when running the tests. I also tested with my generated ukcore code, and confirmed it was working for the new fields there as well.

To require a new field, you'll first need to add it in the module yaml file. In the uscore_v3.1.0_module.yml file, you can see the sequence_requirements section. To add a new requirement, just add one with a label. Then in a sequence file, add that requirement. 

**Submitter:**
- [ ] This pull request describes why these changes were made
- [ ] Internal ticket for this PR:

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/models/module.rb
+++ b/lib/app/models/module.rb
@@ -64,7 +64,7 @@ module Inferno
     end
 
     def variable_required_by(variable)
-      sequences.select { |sequence| (sequence.requires.include? variable) || (sequence.new_requires.include? variable) }
+      sequences.select { |sequence| sequence.requires.include? variable }
     end
 
     def variable_defined_by(variable)
--- a/lib/app/models/module.rb
+++ b/lib/app/models/module.rb
@@ -19,6 +19,7 @@ module Inferno
     attr_accessor :test_sets
     attr_accessor :title
     attr_accessor :resource_path
+    attr_accessor :sequence_requirements
 
     def initialize(params)
       @name = params[:name]
@@ -35,6 +36,13 @@ module Inferno
         params[:test_sets].each do |test_set_key, test_set|
           self.default_test_set ||= test_set_key.to_s
           test_sets[test_set_key] = TestSet.new(test_set_key, test_set)
+        end
+      end
+      if params[:sequence_requirements].present?
+        @sequence_requirements = {}.tap do |requirements|
+          params[:sequence_requirements].each do |requirement_key, texts|
+            requirements[requirement_key.to_sym] = texts
+          end
         end
       end
 
--- a/lib/app/models/module.rb
+++ b/lib/app/models/module.rb
@@ -56,7 +56,7 @@ module Inferno
     end
 
     def variable_required_by(variable)
-      sequences.select { |sequence| sequence.requires.include? variable }
+      sequences.select { |sequence| (sequence.requires.include? variable) || (sequence.new_requires.include? variable) }
     end
 
     def variable_defined_by(variable)
--- a/lib/app/models/sequence_requirement.rb
+++ b/lib/app/models/sequence_requirement.rb
@@ -8,6 +8,7 @@ module Inferno
       property :name, String
       property :value, String
       property :label, String
+      property :description, String
 
       belongs_to :testing_instance
     end
--- a/lib/app/models/sequence_requirement.rb
+++ b/lib/app/models/sequence_requirement.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Inferno
+  module Models
+    class SequenceRequirement
+      include DataMapper::Resource
+      property :id, String, key: true, default: proc { SecureRandom.uuid }
+      property :name, String
+      property :value, String
+      property :label, String
+
+      belongs_to :testing_instance
+    end
+  end
+end
--- a/lib/app/models/sequence_requirement.rb
+++ b/lib/app/models/sequence_requirement.rb
@@ -5,7 +5,8 @@ module Inferno
     class SequenceRequirement
       include DataMapper::Resource
       property :id, String, key: true, default: proc { SecureRandom.uuid }
-      property :name, String
+      property :name, String, unique_index: :name_by_instance
+      property :testing_instance_id, String, unique_index: :name_by_instance
       property :value, String
       property :label, String
       property :description, String
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -279,13 +279,12 @@ module Inferno
         return unless requirement&.dig(:name)
 
         attributes = { name: requirement[:name],
-                       testing_instance_id: id,
                        testing_instance: self,
                        label: requirement[:name],
                        value: '' }
           .merge(requirement)
 
-        condition_attributes = [:name, :testing_instance_id, :testing_instance]
+        condition_attributes = [:name, :testing_instance]
 
         # Get the conditions that determine whether the record already exists
         conditions = attributes.select { |key, _value| condition_attributes.include?(key) }
@@ -302,11 +301,11 @@ module Inferno
       end
 
       def get_requirement_value(requirement_name)
-        get_sequence_requirement(name: requirement_name.to_s, testing_instance_id: id).value
+        get_sequence_requirement(name: requirement_name.to_s, testing_instance: self).value
       end
 
       def set_requirement_value(requirement_name, value)
-        get_sequence_requirement(name: requirement_name.to_s, testing_instance_id: id).update(value: value)
+        get_sequence_requirement(name: requirement_name.to_s, testing_instance: self).update(value: value)
         self.patient_id = value if requirement_name == 'patient_id'
       end
 
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -278,18 +278,12 @@ module Inferno
       def get_sequence_requirement(requirement)
         return unless requirement&.dig(:name)
 
-        attributes = { name: requirement[:name],
-                       testing_instance: self,
+        attributes = { testing_instance: self,
                        label: requirement[:name],
                        value: '' }
           .merge(requirement)
 
-        condition_attributes = [:name, :testing_instance]
-
-        # Get the conditions that determine whether the record already exists
-        conditions = attributes.select { |key, _value| condition_attributes.include?(key) }
-
-        SequenceRequirement.first_or_create(conditions, attributes)
+        SequenceRequirement.first_or_create({ name: requirement[:name], testing_instance: self }, attributes)
       end
 
       def add_sequence_requirements(requirements)
--- a/lib/app/models/testing_instance.rb
+++ b/lib/app/models/testing_instance.rb
@@ -280,13 +280,16 @@ module Inferno
         token_retrieved_at + token_expires_in.seconds
       end
 
-      def add_sequence_requirement(requirement)
-        new_requirement = SequenceRequirement.new(
-          name: requirement,
-          value: '',
-          label: requirement
-        )
-        sequence_requirements.push(new_requirement)
+      def add_sequence_requirements(requirements)
+        requirements.each do |requirement, texts|
+          new_requirement = SequenceRequirement.new(
+            name: requirement,
+            value: '',
+            label: texts[:label],
+            description: texts[:description]
+          )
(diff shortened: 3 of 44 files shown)


radamson:
I think adding the labels and descriptions in the module.yml is a good way to start, but I'm worried about having to manage and update the prerequisites in multiple places.  Right now we have a situation where they need to be edited in the ERBs, model and test code and its a bear to deal with.

I'm not sure if its the best way to do it, but it might be nice to be able to add the label and description when defining the prerequisite so that it all happens in the same place.  Otherwise I could see the prerequisites getting out of sync where maybe we forget to add a label in the config or have a label in the config, but never use it.

I don't want perfect to get in the way of good, but I think longer term we might want to be able to centralize the management of prereqs.

lib/app/sequence_base.rb
```diff
@@ -253,6 +254,32 @@ def self.requires(*requires)
         @@requires[sequence_name] || []
       end
 
+      def self.new_requires(*requires)
```
radamson, line 257: I vaguely remember some conversation on not adding `new_requires`, but instead updating `requires` to be able to handle when the requirement does not already exist.  the logic of this method could still be encapsulated in its own method that `requires` uses to make it easy the logic easier to follow.

I think that just updating `requires` would ultimately be a cleaner and less confusing solution for test writers.  Any big obstacles to that I'm not thinking of?

arscan: > I think adding the labels and descriptions in the module.yml is a good way to start, but I'm worried about having to manage and update the prerequisites in multiple places. Right now we have a situation where they need to be edited in the ERBs, model and test code and its a bear to deal with.

Could take care of this using ruby.  By default have these just be defined completely in the sequences.  And if you use in multiple sequences place in a common ruby file / module and include the common file.  And/or you could have it be smart enough that only 1 sequence has to include the label/description portions, and in the other just require the parameter name it is good enough.

arscan:
> I think adding the labels and descriptions in the module.yml is a good way to start, but I'm worried about having to manage and update the prerequisites in multiple places. Right now we have a situation where they need to be edited in the ERBs, model and test code and its a bear to deal with.

Could take care of this using ruby. By default have these just be defined completely in the sequences. And if you use in multiple sequences place in a common ruby file / module and include the common file. And/or you could have it be smart enough that only 1 sequence has to include the label/description portions, and in the other just require the parameter name it is good enough.

lib/app/sequence_base.rb
arscan: We need to be optimizing for test writers.  I don't quite understand what 'new_requires' means by looking at it.  Is it stuff that is in only my module, and not defined in another module?

I personally think we should just consider things in 'requires' as 'global variables' that can be shared between modules.  If we want to be a little more explicit about where the variable is defined, we could add some kind of namespacing, so that if the SMART sequence defines "token" maybe to refer to it elsewhere you do `:requires SMART.token` or something.

radamson: I think the namespacing could be cool.  Since we currently expect the requires to be accessible as methods on `@instance` though it might be a little tricky so I'd lean towards making that change in a future PR so we can at least have the baseline functionality in quick.

czh-orz: Yeah, I should have explained in the PR, but `new_requires` is just a placeholder to show how the pattern would look without having to go back and change all the previous requires. Ultimately, it would just use `requires`

lib/app/models/testing_instance.rb
```diff
@@ -182,14 +180,15 @@ def patient_id=(patient_id)
...
 
         # For patient id list, don't clear it out but rather add it to the list of known
         # patients to pull from.
-        self.patient_ids = if patient_ids.blank?
-                             patient_id
-                           else
-                             patient_ids.split(',').append(patient_id).uniq.join(',')
-                           end
+        if patient_ids.blank?
+          set_requirement_value('patient_ids', patient_id)
```
Jammjammjamm, line 188: You should be able to use the getter/setter here.

lib/app/sequence_base.rb
```diff
@@ -250,6 +250,27 @@ def self.details(details = nil)
...
+          instance_class.define_method requirement_name do
+            requirement_found = sequence_requirements.find { |requirement| requirement.name == requirement_name.to_s }
+            return unless requirement_found.present?
+
+            requirement_found.value
+          end
+
+          instance_class.define_method requirement_setter_name do
+            requirement_found = sequence_requirements.find { |requirement| requirement.name == requirement_name.to_s }
+            return unless requirement_found.present?
```
Jammjammjamm, line 268: Should this create the requirement if it isn't found?

lib/app/models/sequence_requirement.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Inferno
+  module Models
+    class SequenceRequirement
+      include DataMapper::Resource
+      property :id, String, key: true, default: proc { SecureRandom.uuid }
+      property :name, String
```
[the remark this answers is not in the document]
czh-orz: Added, though not sure if I'm supposed to pass in the testing_instance_id when creating or if it'll figure it out.

lib/app/models/testing_instance.rb
```diff
@@ -182,14 +180,15 @@ def patient_id=(patient_id)
         return if patient_id.to_s == self.patient_id.to_s
 
         resource_references.destroy
+        patient_ids = get_requirement_value('patient_ids')
```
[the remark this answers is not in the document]
czh-orz: Yeah, it's just an example to show that the method works without adding any new fields.

lib/app/sequence_base.rb
```diff
@@ -250,6 +250,27 @@ def self.details(details = nil)
...
       def self.requires(*requires)
         @@requires[sequence_name] = requires unless requires.empty?
+
+        instance_class = Inferno::Models::TestingInstance
+        requires.each do |requirement_name|
+          requirement_setter_name = "#{requirement_name}=".to_sym
+          next if instance_class.method_defined?(requirement_name) && instance_class.method_defined?(requirement_setter_name)
+
+          instance_class.define_method requirement_name do
+            requirement_found = sequence_requirements.find { |requirement| requirement.name == requirement_name.to_s }
```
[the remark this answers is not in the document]
czh-orz: fixed

lib/app/models/testing_instance.rb
czh-orz: yep, fixed

lib/app/models/sequence_requirement.rb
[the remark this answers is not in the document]
Jammjammjamm: In `add_sequence_requirements`, rather than a `new` and a `push`, use a create and pass in the testing instance.
```ruby
requirements.each do |requirement, texts|
  SequenceRequirement.create(
    name: requirement,
    value: '',
    label: texts[:label],
    description: texts[:description],
    testing_instance: self
    # "testing_instance_id: id" should work if this doesn't
  )
end
```

You also shouldn't need the final `save!` since the `TestingInstance` itself isn't being changed.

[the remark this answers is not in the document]
czh-orz: I think this is working with add_sequence_requirements, but its causing errors when I try to add requirements in get_requirement_value and set_requirement_value. Trying it this way and running rake gives a ton of errors.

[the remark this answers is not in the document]
radamson: I was also getting similar errors around the unique constraints on `name` and `testing_instance_id`

e.g. `DataObjects::IntegrityError: UNIQUE constraint failed: inferno_models_sequence_requirements.name, inferno_models_sequence_requirements.testing_instance_id`

I was able to resolve it by adding another check before creating and using `#first`.  Interestingly using either:

`sequence_requirements.find { |req| req.name == requirement_name.to_s }`
or
`sequence_requirements.find { |req| req.name == requirement_name.to_s && req.testing_instance_id == id }`

was not returning up to date results, while

`sequence_requirements.first(name: requirement_name.to_s, testing_instance_id: id)` was (you can test by just swapping back the existing).

This is leading me to believe that using `#find` might not be the most reliable way to search through the associated tables (maybe there is an "update references" or something method, but I think using `#first` seems clearer and its a DataMapper method built for this purpose).


Either way seems to be working now!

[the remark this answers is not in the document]
Jammjammjamm: These errors are why we needed the uniqueness constraint. We don't want multiple instances of these values to exist. Using `.first` and then creating or updating introduces a race condition (the record could be created in between when you check and when you create).

`DataMapper` has a [`first_or_create` method](https://www.rubydoc.info/github/datamapper/dm-core/DataMapper%2FModel:first_or_create) that should prevent this race condition from occurring. Here's [an example using it](https://stackoverflow.com/questions/6509788/datamapper-first-or-create-always-set-certain-field).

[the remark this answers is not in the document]
radamson: Awesome, good suggestion!  Changed to use `#first_or_create`.  Added a helper method around it to contain the defaults and logic for conditions and defaults logic.

lib/app/models/testing_instance.rb
```diff
@@ -281,6 +275,40 @@ def token_expiration_time
...
+                       label: requirement[:name],
+                       value: '' }
+          .merge(requirement)
+
+        condition_attributes = [:name, :testing_instance]
+
+        # Get the conditions that determine whether the record already exists
+        conditions = attributes.select { |key, _value| condition_attributes.include?(key) }
+
+        SequenceRequirement.first_or_create(conditions, attributes)
```
Jammjammjamm, line 292: I think lines 287-290 are overcomplicating things.
```suggestion
        SequenceRequirement.first_or_create({ name: requirement[:name], testing_instance: self }, attributes)
```

lib/app/models/testing_instance.rb
```diff
@@ -281,6 +275,40 @@ def token_expiration_time
         token_retrieved_at + token_expires_in.seconds
       end
 
+      def get_sequence_requirement(requirement)
+        return unless requirement&.dig(:name)
+
+        attributes = { name: requirement[:name],
```
Jammjammjamm, line 281: Don't need to set `name: requirement[:name]` if we're then merging in `requirement`.

lib/app/models/module.rb
```diff
@@ -37,6 +38,13 @@ def initialize(params)
           test_sets[test_set_key] = TestSet.new(test_set_key, test_set)
         end
       end
+      if params[:sequence_requirements].present?
+        @sequence_requirements = {}.tap do |requirements|
+          params[:sequence_requirements].each do |requirement_key, texts|
```
Jammjammjamm, line 43: Why is this value called `texts`? The same thing is done in `TestingInstance`.

lib/app/models/testing_instance.rb
```diff
@@ -281,6 +275,40 @@ def token_expiration_time
...
+        end
+      end
+
+      def get_requirement_value(requirement_name)
+        get_sequence_requirement(name: requirement_name.to_s, testing_instance: self).value
+      end
+
+      def set_requirement_value(requirement_name, value)
+        get_sequence_requirement(name: requirement_name.to_s, testing_instance: self).update(value: value)
+        self.patient_id = value if requirement_name == 'patient_id'
```
Jammjammjamm, line 309: I hope we're not keeping this in the final version.

czh-orz: Took this out. Forgot we have the setters now so I'm back to using that in the test_set_endpoint

lib/app/models/module.rb
czh-orz: changed name to `req_description`.

arscan:
Not a blocking issue, but I was very specific about how we ordered the fields that come up in the form.   At least now, the order is not quite right -- though that may be because this is a mix of the old and new methods... perhaps if we went to all the new method they will show up in the interface in the order that they are put in the module file?

PR #471, opened 2020-07-20
FI-926: Load IGs from packages.fhir.org

jason-crowley:
This PR adds two new methods: `load_ig_by_id` and `load_ig_by_tgz`, which will allow IGs from packages.fhir.org to be loaded into the `HL7Validator` by their NPM ID or by a `package.tgz`, respectively. This PR requires [`fhir-validator-wrapper/load-user-packages`](https://github.com/inferno-community/fhir-validator-wrapper/pull/16) to work.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-926

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/app/utils/hl7_validator.rb
```diff
@@ -25,10 +27,31 @@ def validate(resource, fhir_models_klass, profile_url = nil)
       information = issues_by_severity(outcome.issue, 'information')
       {
         errors: fatals.concat(errors).reject(&:empty?),
-        warnings: warnings.concat(information).reject(&:empty?)
+        warnings: warnings,
+        information: information
```
Jammjammjamm, line 31: I think we were intentionally not including informational messages here.

radamson: I think we do want them here and this will make the functionality closer to the [recent code in program](https://github.com/onc-healthit/inferno-program/blob/b9779b708e2262dd21f129bbf7b9f8740e0e3d69/lib/app/utils/hl7_validator.rb#L29) where information messages are not lumped in with warnings.

I don't think they are actually used anywhere, but adding them here at least makes them available to tests and we won't be throwing warnings unexpectedly.

I vaguely remember there was a PR somewhere to use information messages and it included a little `i` icon, but I can't find it.

radamson: Could be wrong though

okeefm: In `inferno-program`, we do support the gathering of `@information_messages` in `sequence_base.rb`, and we add any `information` messages from the validator into that array.

okeefm: I can't actually find where we display them, though.

Jammjammjamm: I think we removed them because we decided they weren't something we wanted to display in program.

radamson: here was the PR with the little `i`.  Looks like in the last commit we decided not to display the informational messages separately.

I still think this change is worthwhile though so that we don't show information messages as warnings.

test/test_helper.rb
```diff
@@ -56,6 +56,11 @@ def load_fixture_with_extension(file_name)
   File.read(File.join(fixture_path, file_name))
 end
 
+def load_binary_fixture(file_name)
+  fixture_path = find_fixture_directory
+  File.open(File.join(fixture_path, file_name), 'rb', &:read)
+end
```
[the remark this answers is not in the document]
Jammjammjamm: i think this is fine as is.

lib/app/utils/hl7_validator.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'json'
```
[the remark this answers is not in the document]
radamson: we use the json library further down with `JSON.parse(result.body)`.  Being explicit makes sure this file always has that loaded...we've ran into problems previously where we relied on other files to do the requiring for us, but then this file could break if the other file ever removes that require.  Plus being explicit is nice IMO.

okeefm:
OBE: Covered by #476 as we're not pulling from packages.fhir.org currently in inferno-community, but we may revisit when we have time to implement that.

PR #477, opened 2020-08-04
Fi 898 generic generator from capability statement

czh-orz:
This PR adds search and interaction tests based off the capability statement for the generic generator. The search tests only perform the basic search and validate functionality. It doesn't handle comparators and modifiers yet. 

It also defines a new generic_utilities file that is included in all the generated sequences. You can add functions that all the generated sequences can access through that file. This way, we don' need to modify sequence_base to create shared code. 

It also adds the sequence definition files like there is in us core. Right now, it contains search parameter metadata, but it will later contain the terminology and must support information. 

I also added a generic_generator_utilities file that is used to store shared code across the generator.


generator/generic/static/generic_utilities.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+module Inferno
+  module GenericUtilities
```
Jammjammjamm, line 4: We should give this a name that makes more sense when it's included in a sequence. `SequenceUtilities`? `SequenceHelpers`?

Having it located in `static` is a problem because it gets copied into each generated module, but this file should only exist and be loaded once. If it's going to exist with each module, then it needs to be namespaced.

generator/sequence_metadata.rb
```diff
@@ -51,10 +69,69 @@ def url
...
+        end
+        search_combinations = search_combinations_from_capability_statement(capabilities)
+        searches.append(search_combinations) unless search_combinations.nil?
+        searches
+      end
+
+      def search_combinations_from_capability_statement(capabilities)
+        search_combo_url = 'http://hl7.org/fhir/StructureDefinition/capabilitystatement-search-parameter-combination'
+        capabilities['extension']
+            &.select { |ext| ext['url'] == search_combo_url }
```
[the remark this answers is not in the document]
radamson: I updated the spacing and removed the redundant variable.

The spacing is still a little wonky but Rubocop wants it this way 🤷‍♂️ .

generator/generic/static/generic_utilities.rb
radamson: Renamed it to `GenericUtilities` and moved it into `lib/utils`.  Sequence base requires the file, but the individual sequences are responsible for including it.

generator/sequence_metadata.rb
```diff
@@ -51,10 +69,69 @@ def url
...
+
+      def searches
+        @searches ||= searches_from_capability_statement(capabilities)
+      end
+
+      def searches_from_capability_statement(capabilities)
+        return [] unless capabilities.present?
+        searches = []
+        basic_searches = capabilities['searchParam']
+        basic_searches&.each do |search_param|
```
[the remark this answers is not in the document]
radamson: Removed capabilities from method arguments and use `attr_reader` instead.

generator/generic/templates/sequence_definition.rb.erb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Inferno
+  module USCore310ProfileDefinitions
```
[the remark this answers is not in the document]
radamson: @czh-orz renamed this to `ProfileDefinitions`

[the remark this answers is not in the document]
Jammjammjamm: It can't just be `ProfileDefinitions` because then every generated sequence will share the same definitions. It needs to be `Inferno::SOMETHING_UNIQUEProfileDefinitions` or `Inferno::SOMETHING_UNIQUE::ProfileDefinitions`.

generator/generic_generator_utilities.rb
```diff
@@ -0,0 +1,140 @@
+# frozen_string_literal: true
+
+module Inferno
+  module Generator
+    module GenericGeneratorUtilties
```
Jammjammjamm, line 5: I don't think this module forms a cohesive unit.

- `structure_to_string` is only used by `GenericGenerator`
- `create_search_validation` is only used by `SequenceMetadata`
- the other methods are helpers for `create_search_validation`

The name doesn't help you know why a method would be in here rather than in one of those classes.

generator/generator_base.rb
```diff
@@ -27,7 +27,13 @@ def load_resources
 
             # We should consider using the native Ruby models instead of JSON
             # There were problems with round-tripping certain SearchParameters though
-            new_resource_json = JSON.parse(File.read(resource))
+
+            begin
+              new_resource_json = JSON.parse(File.read(resource))
+            rescue JSON::ParserError # failed to parse json
+              next
```
Jammjammjamm, line 34: Should probably log something here.

generator/generic_generator_utilities.rb
```diff
@@ -0,0 +1,140 @@
...
+                case property
+                  #{search_validators}
+                end
+              end
+            )
+          end
+        validate_functions
+      end
+
+      def search_param_match_found_code(type, element)
```
Jammjammjamm, line 65: Rather than duplicating this code in each sequence, can you stick this method's logic (without all the stringifying) into `SequenceUtilities`.

generator/generic/templates/sequence_definition.rb.erb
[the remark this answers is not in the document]
radamson: Renamed so it includes the `path`.  If the module is in the `ig` directory it will be named `IgProfileDefinitions` and if its in the `uscore` directory it will be `UscoreProfileDefinition`.

generator/sequence_metadata.rb
```diff
@@ -1,21 +1,39 @@
...
+      def initialize(profile, all_search_parameter_metadata, capability_statement = nil)
         @profile = profile
         @tests = []
+        return unless capability_statement.present?
+
+        @capabilities = capability_statement['rest']
+          .find { |rest| rest['mode'] == 'server' }['resource']
+          .find { |resource| resource['type'] == profile['type'] }
+
+        @search_parameter_metadata = capabilities['searchParam']&.map do |param|
```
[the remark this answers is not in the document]
radamson: Agreed, but I think "near future" needs to be a future PR 🙂 .

generator/sequence_metadata.rb
```diff
@@ -51,16 +73,74 @@ def url
...
+      end
+
+      def search_combinations_from_capability_statement
+        return [] unless capabilities.present?
+
+        search_combo_url = 'http://hl7.org/fhir/StructureDefinition/capabilitystatement-search-parameter-combination'
+        capabilities['extension']
+          &.select { |ext| ext['url'] == search_combo_url }
+          &.map do |combo|
+          expectation = combo['extension'].find { |ext| ext['url'] == EXPECTATION_URL }['valueCode']
```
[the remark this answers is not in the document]
radamson: As with the previous comment on this, rubocop has a problem with me fixing the indentation of the code in the `do` block.

I tried updating rubocop to see if that fixed it incase a recent update (like https://github.com/rubocop-hq/rubocop/pull/6749) had changed this, but no 🎲 .

Instead I added disabled the rubocop indentation checks for just that block of code.  Hilariously, I also had to disable the CommentIndentation check as well to indent the `rubocop:disable` directive the way I wanted.

generator/generator_base.rb
czh-orz: 👍

generator/generic_generator_utilities.rb
czh-orz: I think this is reasonable. But I do remember Rob saying something about writing out functions explicitly like this in each sequence to make it easier to understand. But anyway, can we leave this to a future PR?

generator/generic_generator_utilities.rb
czh-orz: I wanted to leave the `create_search_validation` stuff out of `SequenceMetadata` because I wanted to keep that file for just profile information.  `create_search_validation` just uses the metadata to form code and doesn't add any metadata. I couldn't put it in `search_test` either because it doesn't go inside of a test itself. 

`structure_to_string` is going to be needed later on when I add must support stuff.

generator/generic_generator_utilities.rb
arscan: My preference in general is for it to be fairly obvious what a given test does without having to continually hop to other shared code locations.  Yeah, it results in duplication of generated code, but that duplicated code never needs to managed indivdually (its managed as a group in the generator).  Just a preference.  But if we end up with a lot of tests that basically are just a call to a shared function that needs to be followed to understand what is going on at all i think we are doing it wrong.

PR #484, opened 2020-08-18
fi-949 merging program us core to community

czh-orz:
This PR merges us core tests from the program edition to the community version, with the exception of the profile validation tests which will be merged later. For the most part I just copied over the entire us core generator folder and added whatever other changes needed to make this work. These other changes are in the model files such as request_response.

To test, run bundle exec rake inferno:generate['uscore','uscore_v3.1.0']


generator/uscore/static_test/access_verify_restricted_test.rb
```diff
@@ -0,0 +1,208 @@
+# frozen_string_literal: true
+
+# NOTE: This is a generated file. Any changes made to this file will be
+#       overwritten when it is regenerated
+
+require_relative '../../../../test/test_helper'
+
+describe Inferno::Sequence::ONCAccessVerifyRestrictedSequence do
```
Jammjammjamm, line 8: Do we want these program-specific sequences in community?

generator/uscore/templates/sequence.rb.erb
```diff
@@ -1,31 +1,99 @@
...
+        are validated against this profile.<% end %>
+
+        # Testing Methodology
+
+        <% if resource == 'Provenance' %>
+        Previously run sequences store references to US Core Provenance resources that are associated with other US Core
+        resources using the appropriate `_revincludes` search.  This set of tests uses these found resources to verify
+        support for the `read` operation.  Each of these resources must conform to the US Core Provenance profile.
+        <% elsif delayed_sequence %>
+        Because <%=resource%> resources are not present o not exist in USCDI, no searches are performed on this test sequence. Instead, references to
```
Jammjammjamm, line 33: > present o not
typo

Also, do we want these tests to be specific to US Core rather than the program requirements?

generator/uscore/uscore_generator.rb
```diff
@@ -730,13 +700,15 @@ def structure_to_string(struct)
...
-          tests_that: "#{sequence[:resource]} resources returned conform to US Core R4 profiles",
+          tests_that: "#{sequence[:resource]} resources returned from previous search conform to the #{sequence[:profile_name]}.",
           key: test_key,
           index: sequence[:tests].length + 1,
           link: sequence[:profile],
           description: %(
-            This test checks if the resources returned from prior searches conform to the US Core profiles.
-            This includes checking for missing data elements and valueset verification.
+            This test verifies resources returned from the first search conform to the [US Core #{sequence[:resource]} Profile](#{sequence[:profile]}).
+            It verifies the presence of manditory elements and that elements with required bindgings contain appropriate values.
```
Jammjammjamm, line 709:
```suggestion
            It verifies the presence of mandatory elements and that elements with required bindings contain appropriate values.
```

generator/uscore/uscore_generator.rb
```diff
@@ -1217,10 +1287,52 @@ def test_medication_inclusion(medication_requests, search_params)
...
             medications = requests_with_medications.select { |resource| resource.resourceType == 'Medication' }
             assert medications.present?, 'No Medications were included in the search results'
+
+            @medications += medications
+            @medications.uniq!(&:id)
           end
         )
       end
 
+      def generate_verify_access_module(module_info)
```
Jammjammjamm, line 1297: This again seems like a program-specific test rather than a US Core test.

czh-orz:
> Could you update the 'Clinical Notes Guidance Tests' because it expects a 'patient_id' instead of 'patient_ids'. That causes the following kinda weird prompt when you do 'run all':
> 
> Screen Shot 2020-08-21 at 12 48 28 PM
> 
> I also think our Must Support checking test at the bottom is a little confusing in the context, because you have to run others first (without much of an explanation that you need to do that). It would be some work to be more explicit (by having a param of Patient/1, Encounter/5 that it will use to find examples), but is something I feel like we should do eventually. This speaks to general test design considerations that I think we can get better at, but don't think we should fix right now.

Yep, should be fixed now

generator/uscore/static_test/access_verify_restricted_test.rb
czh-orz: Removed access verify tests

generator/uscore/uscore_generator.rb
czh-orz: Removed

Jammjammjamm:
These kinds of errors will be resolved when we do the terminology update, right?

Screen Shot 2020-09-08 at 1 52 03 PM

arscan (approved):

arscan:
Are we good here?  SHould we open another ticket to verify our terminology setup in community instead of addressing it now?  Unfortunately we need this in now so we can get 3.1.1 in to start testing with.

arscan:
I created a ticket to investigate our terminology setup in community because this is blocking other work now.  If that's it we should merge this in now.

PR #487, opened 2020-08-25
FI-942 BDT update

yunwwang:
Update BDT service to the latest Aug. 11th commit

### DIFF
--- a/generator/bdt/bdt-structure.json
+++ b/generator/bdt/bdt-structure.json
@@ -20,17 +20,10 @@
             },
             {
               "id": "Auth-01.0.1",
-              "name": "Rejects expired token",
-              "description": "The server should reject expired tokens at the system-level export endpoint",
-              "type": "test",
-              "path": "0.0.1"
-            },
-            {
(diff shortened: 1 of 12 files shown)


Jammjammjamm:
Should these all pass against the bulk data preset? I get a few errors:

Screen Shot 2020-09-08 at 1 40 03 PM

Screen Shot 2020-09-08 at 1 40 39 PM

arscan:
I was able to update the BDT service, build it, and point this branch to it, and it seemed to work.  There were some failures against the server I was testing, but I feel like those should be addressed separately.

Also, I noticed the interface lists the requests/results backwards (that may be due to a bug that we addressed regarding request/response orders, that may not have landed in community, or may need to be updated slightly in this code?)

arscan:
Turns our the request/response log is backwards because a fix we brought to program edition didn't make it to community. I created a jira ticket for that, I don't think we need to address here.

arscan (approved):
I'm good with this.  It appears the integration is working now, and we can address individual test issues separately.

PR #491, opened 2020-09-04
Add demo server preset for local development

Jammjammjamm:


arscan (approved):
Worked fine for me!

okeefm (approved):
Tested in local docker (accessed via `http://localhost:4567`) and all these presets worked. Looks good!

PR #495, opened 2020-09-08
FI-998 USCore 3.1.1

yunwwang:
Update Community to USCore 3.1.1 including:
* US Core 3.1.1 profiles
* Regenerate Test Sequences
** Pediatric profiles with additional must_support from base profile (FHIR-27946)
** Remove must support to Non US Core profiles (FHIR-27876)
** Add reaction to US Core AllergyIntolerance (FHIR-27867)
** Change reference to US Core Patient (FHIR-27857)
** Add US core Head Occipital-frontal Circumference Percentile Profile  (FHIR-27542)
** Remove must supported CLIA from US Core Organization (FHIR-27158)
* Remove special logic to generate 'ge' operator (FHIR-27893)
* Update unit test for multi-or
* Add USCore version to generator template


generator/uscore/templates/sequence.rb.erb
```diff
@@ -7,7 +7,7 @@ module Inferno
   module Sequence
     class <%=class_name%> < SequenceBase
       include Inferno::DataAbsentReasonChecker  
-      include Inferno::USCore310ProfileDefinitions
```
arscan, line 10: I'm confused why you are making this version independent.  I'd think the profile definitions are version specific?

arscan (changes requested):

generator/uscore/templates/sequence.rb.erb
yunwwang: The class is version dependent while module should not be. 
image

yunwwang: I should regenerate v3.1.0 tests too.

Jammjammjamm: What was the reason for this change? `USCore311CarePlanSequenceDefinitions` will never be instantiated, so it makes sense for it to be a module that gets included rather than a class living in a module that gets included.

arscan: Oh ok.  Yes it is critical that this does not break US Core v3.1.0 tests if you regenerate them (or not).  And we need to get whatever you've changed in here back to Program so we don't fall out of synch.

arscan: If making the generator generic between the two is too big a lift for today, we could merge in just the tests to this release and leave your generator changes in a separate branch to be completed later.

yunwwang: @arscan Yes. Remove version from module name is critical to allow different US Core version lives side by side. 
@Jammjammjamm I am trying to minimize changes while adopting 3.1.1 so the Profile Definition module is based on existing 3.1.0 branch. I can create a JIRA ticket to optimize this area.

yunwwang: Created https://oncprojectracking.healthit.gov/support/browse/FI-1004 for generating SequenceDefinition as a module.

arscan: We fixed this to keep the same naming convention.

arscan (approved):

PR #496, opened 2020-09-09
Move values_found to inside the patients loop, so we get at least one patient with 2 status values

okeefm:
This PR moves the values_found counter in the US Core search tests into the patient loop, so we look for at least two statuses for each resource type for each patient. This is needed to allow for the multipleOr tests that occur later on in the test suites.

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -1013,7 +1013,7 @@ module Inferno
           #{skip_if_search_not_supported_code(sequence, search_parameters)}
           @#{sequence[:resource].underscore}_ary = {}
           @resources_found = false
-          validated_search_param_variants = false
+          search_query_variants_tested_once = false
           #{values_variable_name} = [#{search_param[:values].map { |val| "'#{val}'" }.join(', ')}]
           patient_ids.each do |patient|
             @#{sequence[:resource].underscore}_ary[patient] = []
@@ -1035,9 +1035,7 @@ module Inferno
               save_delayed_sequence_references(resources_returned, #{sequence[:class_name]}Definitions::DELAYED_REFERENCES)
               validate_reply_entries(resources_returned, search_params)
 
-              #{'test_medication_inclusion(@medication_request_ary[patient], search_params)' if sequence[:resource] == 'MedicationRequest'}
-
-              next if validated_search_param_variants
+              next if search_query_variants_tested_once
               #{get_token_system_search_code(search_parameters, sequence)}
 
               search_params_with_type = search_params.merge('patient': "Patient/\#{patient}")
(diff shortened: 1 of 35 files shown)


czh-orz (approved):

generator/uscore/uscore_generator.rb
```diff
@@ -1024,9 +1024,9 @@ def get_first_search_with_fixed_values(sequence, search_parameters, save_resourc
           #{skip_if_search_not_supported_code(sequence, search_parameters)}
           @#{sequence[:resource].underscore}_ary = {}
           @resources_found = false
-          #{'values_found = 0' if find_two_values}
           #{values_variable_name} = [#{search_param[:values].map { |val| "'#{val}'" }.join(', ')}]
           patient_ids.each do |patient|
+            #{'values_found = 0' if find_two_values}
```
Jammjammjamm, line 1029: There's only a `break` in the inner `.each`, so even when a patient is found with two status, it continues to perform searches for each other patient. Is that the intention?

Screen Shot 2020-10-05 at 2 57 01 PM

arscan: I think this always worked this way, and this doesn't change that. @czh-orz should weigh in.

arscan: I think in this case, since this is the most 'general' of careteam searches, we shouldn't even be stopping after we find 2 different statuses within the same patient, because we might not find all resources that are relevant for hte 'must support' checks later.  We are communicating (or at least I am), that you must have all the must support elements somewhere across all resources, but we stop prematurely in this case it seems?

czh-orz: You're right that this might affect the must support checks. In that case, we shouldn't break once two have been found and we should just loop through all the patients provided.

arscan:
I updated to ensure the first search always gets all possible resources (which is needed for the must support checks at the end).  I also added a little code to reduce redundant searches within the patient_id, code inner loop so only in the first loop it checks things like Patient/id and code|system variants.  Please take a look @czh-orz @okeefm @Jammjammjamm

Jammjammjamm:
The CareTeam test looks like it's working as intended, but the MedicationRequest test didn't get through all of the intents:

Screen Shot 2020-10-07 at 8 19 56 AM

I think this is because `validate_reply_entries` will raise an assertion error before the test has finished going through all of the values.

arscan:
Was it a valid assertion failure @Jammjammjamm?  If so, I feel like its fine if it doesn't actually make all the queries, the end result is the same.

Jammjammjamm:
Yeah, I think it's one of the errors that will be addressed by #502

Jammjammjamm (approved):

arscan (approved):

PR #519, opened 2020-12-22
FI-1100 Migrate USCore 311 code generation update to Community

yunwwang:
This PR includes
* Update code generator and templates for US Core IG v3.1.1 
* Remove testing for US Core IG 3.1.0

This PR does not include:
* Validator update for US Core IG. Validate update is include in another PR


generator/uscore/metadata_extractor.rb
```diff
@@ -556,6 +549,10 @@ def set_first_search(sequence, params)
         sequence[:searches].delete(search)
         sequence[:searches].unshift(search)
       end
+
+      def capitalize_first_letter(str)
```
Jammjammjamm, line 553: We should get rid of this method and go back to using `String#upcase_first`, which we get from `ActiveSupport`.

Jammjammjamm (commented):
When you say the validator update will be in a separate PR, what do you mean by that? The changes in these PRs?
- https://github.com/onc-healthit/inferno-program/pull/212
- https://github.com/onc-healthit/inferno-program/pull/214

yunwwang:
This validator PR: https://github.com/onc-healthit/inferno/pull/502

generator/uscore/metadata_extractor.rb
yunwwang: changes are reversed. Why do we have to use different pattern in Community?

arscan:
Replaced by #526

PR #530, opened 2021-01-19
IPS Tests

radamson:
Adds the IPS tests that were developed for Connectathon 26.  Includes read and validate tests for profiles along with a test for `$document`.

_**Note: This PR will be merged directly into `master`, not `development`**_


arscan (approved):
I was able to get it run locally and the tests appear to work in the case where we do not have any real data.  The surface area changed outside of the tests (and supporting profiles) is limited, so I think this is good to merge.

PR #533, opened 2021-01-22
Update version file.

arscan:
Missed updating the version file in v2.12.0.  Since this is community, and it is only really 'informational' for the user, I don't think it is a huge deal that our release doesn't reflect it, but if there is an easy way to fix the tags we can give it a shot (probably not).  We could alternatively do a quick v2.12.1 release that updates it again?

### DIFF
--- a/lib/version.rb
+++ b/lib/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Inferno
-  VERSION = '2.11.0'
+  VERSION = '2.12.0'
 end


radamson (approved):

radamson:
It would be nice if we could update the release. It looks like we could update the tag locally and push it, but I'm not sure if the source assets would update automatically.

Easiest might be to just delete the release and create it fresh.

arscan:
Will deleting the release delete the tag too?  I figured the release just points to a tag.  Deleting tags is not something I understand very well so It makes me nervous to do anything with it.

PR #535, opened 2021-02-05
Remove unnecessary files, update image name, add batch.

arscan:
This PR:

* Removes a bunch of old files that no longer apply
* Renames 'ruby_server' to 'inferno' in the docker-compose file, to be consistent with program
* Adds a 'batch' directory and volume so that we have somewhere to put the CI 'batch' files somewhere
* Adds a 'batch' file that works
* Persists the data directory in a volume in the docker-compose

I primarily did this to make the batch scripting work again, but then cleaned up a few things that have been around for a long time.


Jammjammjamm:
https://github.com/onc-healthit/inferno/blob/main/deployment-files/nginx.conf#L73 needs updating

Jammjammjamm (approved):

okeefm (approved):
Looks good

PR #543, opened 2021-04-05
Smart Scheduling Links tests

arscan:
Work in progress.  Will use this PR to manage remaining tasks.

You can use with the example URL: https://raw.githubusercontent.com/smart-on-fhir/smart-scheduling-links/master/examples


arscan:
Merging despite code climate rubocop issues.  We believe it has to do with strangeness around branches.  But we cannot hold up this branch any longer because debugging the external tool has become the bottleneck.

okeefm (approved):
Tested this locally, works for me as expected at this point. Rubocop all passes locally.

PR #546, opened 2021-04-08
Smart scheduling test cases

hershilpatel:
Added initial batch of test cases for scheduling.  Some tests are commented out at the moment.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,7 +8,7 @@ checks:
   complex-logic:
     enabled: true
   file-lines:
-    enabled: false
+    enabled: true
   method-complexity:
     enabled: true
   method-count:
@@ -29,3 +29,5 @@ plugins:
     channel: rubocop-0-67
     config:
       file: .rubocop_todo.yml
+exclude_patterns:
+- "**/test/"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,8 +1,7 @@
 version: "2"         # required to adjust maintainability checks
 exclude_patterns:
-  - "!**/test/"
+  - "!**/test/*"
   - "!db/**/*"
-  - "!lib/modules/smart_scheduling_links/test/*"
 checks:
   argument-count:
     enabled: true
@@ -26,7 +25,7 @@ checks:
     enabled: true
 plugins:
   rubocop:
-    enabled: false
-    channel: rubocop-0-67
+    enabled: true
+    channel: rubocop-0-80
     config:
       file: .rubocop.yml
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,7 +8,7 @@ checks:
   complex-logic:
     enabled: true
   file-lines:
-    enabled: true
+    enabled: false
   method-complexity:
     enabled: true
   method-count:
@@ -28,4 +28,4 @@ plugins:
     enabled: false
     channel: rubocop-0-67
     config:
-      file: .rubocop.yml
+      file: .rubocop_todo.yml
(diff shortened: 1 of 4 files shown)


.codeclimate.yml
```diff
@@ -25,7 +25,7 @@ checks:
     enabled: true
 plugins:
   rubocop:
-    enabled: true
-    channel: rubocop-0-80
+    enabled: false
+    channel: rubocop-0-67
     config:
-      file: .rubocop.yml
```
Jammjammjamm, line 31: This doesn't seem right. `rubocop.yml` includes `rubocop_todo.yml`.

hershilpatel: Woops, forgot to undo that change

.codeclimate.yml
```diff
@@ -25,7 +26,7 @@ checks:
     enabled: true
 plugins:
   rubocop:
-    enabled: true
-    channel: rubocop-0-80
+    enabled: false
+    channel: rubocop-0-67
```
arscan, line 30: Why are we reverting back to 0-67 and disabling rubocop?

.codeclimate.yml
```diff
@@ -2,6 +2,7 @@ version: "2"         # required to adjust maintainability checks
 exclude_patterns:
   - "!**/test/"
   - "!db/**/*"
+  - "!lib/modules/smart_scheduling_links/test/*"
```
arscan, line 5: Shouldn't this be covered in the line 2 up from here?  Does that need to be `"!**/test/*"` or is something wrong with your setup?

lib/modules/smart_scheduling_links/test/smart_scheduling_basic_test.rb
```diff
@@ -1,47 +1,1150 @@
...
+
+    @manifest_url = 'http://www.example.com/$bulk-manifest'
+    @location = 'http://www.example.com/locations.ndjson'
+    @locations = [@location]
+    @schedule = 'http://www.example.com/schedules.ndjson'
+    @schedules = [@schedule]
+    @slot = 'http://www.example.com/slots.ndjson'
+    @slots = [@slot]
+
+    @instance = Inferno::TestingInstance.create(url: @base_url, token: @token, selected_module: 'ips')
```
arscan, line 20: This is the wrong module.

lib/modules/smart_scheduling_links/test/smart_scheduling_basic_test.rb
```diff
@@ -1,47 +1,1150 @@
...
+            "type": "Slot",
+            "url": "http://www.example.com/slots-2021-W09.ndjson",
+            "extension": {
+              "state": [
+                "MA"
+              ]
+            }
+          }
+        ]
+      }')
```
arscan, line 226: Something that would cut down a bit on code is put a single "good" Location, Slot and Schedule file in fixtures, load them up in the 'before', then just tweaking them as needed for each failure case.  That would turn 26 lines down into about 1.  Also, for your awareness, you can just use Ruby's literal notation instead of JSON and then parsing the JSON.  But I would recommend moving this fixture into the fixtures directory myself.  Look in another test for an example of loading a fixture.

hershilpatel: looking into it...

.codeclimate.yml
hershilpatel: Undid this change

.codeclimate.yml
hershilpatel: Removed

lib/modules/smart_scheduling_links/test/smart_scheduling_basic_test.rb
hershilpatel: Added fixture

arscan (approved):

Issue #551, opened 2021-04-21
Clinical Notes Guideline Tests does not include patient id in url


AGobron:
**Subject of the issue**
When running a Clinical Notes Guideline Test, the patient ID that is specified as an input does not get included in the url, causing all tests to fail.

**Your environment**
* Edition of inferno (Community or Program): Community
* Version of inferno: 2.11.0
* Which browser and version(s) is the bug present on?: Chrome version 89.0.4389.128

**Steps to reproduce**

1. Go to  https://inferno.healthit.gov/community
2. Click US Core v3.1.1
3. Enter https://inferno.healthit.gov/reference-server/r4 as the server and click begin
4. Click Run for Clinical Notes Guideline Tests
5. Enter SAMPLE_TOKEN as auth token and 85,355 for patient ids. Click Execute
6. Tests will fail. Click 'results...' for the first test. Notice that the patient parameter in the url is just 'patient' and not 'patient=85' or 'patient=355'
7. Click the Inputs tab and confirm that patients 85 and 355 are included in the Patient Ids input
8. Optionally run DocumentReference Tests and compare how the patient parameter in the url should be structured

See https://inferno.healthit.gov/community/f8N9c40PxRb/ for my specific testing results.

**Expected behavior**
A patient ID should be included in the url. The url for the first test should look something like "https://inferno.healthit.gov/reference-server/r4/DocumentReference?patient=85&type=http://loinc.org%7C11488-4"

**Actual behavior**
The "=<patientid>" portion of the url is missing. So the url looks like this: "https://inferno.healthit.gov/reference-server/r4/DocumentReference?patient&type=http://loinc.org%7C11488-4"

yunwwang:
Hello @AGobron : I followed the link you provided. ( https://inferno.healthit.gov/community/f8N9c40PxRb/) The result shows all tests passed and the HTTP Requests logs queries with patient=<patientid> (see attached screen shot). 

![image](https://user-images.githubusercontent.com/8761175/115719641-ea0f3900-a341-11eb-94f8-6ca810f7f1cd.png)

AGobron:
Hi @yunwwang, thanks for looking into this. it looks like your screenshot is of "DocumentReference Tests", which I included in the link for reference. If you scroll down the page you will see the failed "Clinical Notes Guideline Tests".

yunwwang:
Sorry, I looked the wrong sequence. Thank you for reporting the problem to us. We will fix that asap.

[...]

PR #553, opened 2021-04-26
Migrate ClinicalNotes Sequence from Program edition

yunwwang:
# Summary
Fix the issue reported at #551 
Migrated Clinical Notes sequences from Inferno Program to Inferno Community

## Testing guidance
Test using steps mentioned in #551

### DIFF
--- a/lib/modules/us_core_guidance/clinicalnotes_sequence.rb
+++ b/lib/modules/us_core_guidance/clinicalnotes_sequence.rb
@@ -2,16 +2,6 @@
 
 module Inferno
   module Sequence
-    class ClinicalNoteAttachment
-      attr_reader :resource_class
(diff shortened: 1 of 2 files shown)


okeefm (approved):
Clinical Notes sequence looks good to me; identical to program and passes against our reference server.

AGobron:
Hi @arscan and @okeefm, is there a timeframe on when this change will get it's second review?

joeysad2:
@arscan and @okeefm we also would like a timeframe on when this will get a second review and therefore completed. Thanks.

Jammjammjamm (approved):

joeysad2:
@yunwwang do we have a timeline on when this will be merged?

Issue #568, opened 2021-05-24
database contention issues


nathanloyer:
**Subject of the issue**

So I've been trying to reproduce the issue documented [here](https://github.com/onc-healthit/inferno/issues/566) with a vanilla copy of inferno so it is more easily reproducible by anybody. In the process of doing this I uncovered [this other bug with docker-compose](https://github.com/onc-healthit/inferno/issues/567). I was able to produce another error just now that looks related to my original issue (#566), although the error message is not the same. My theory so far is that the ActiveRecord database framework you integrated with earlier this year to replace data mapper is not quite being used correctly when it comes to managing multiple requests in parallel.

**Your environment**
* Edition of inferno (Community or Program): **Community**
* Version of inferno: **main branch**
* Which browser and version(s) is the bug present on?: **Latest chrome as of 05/24/2021**

**Steps to reproduce**

I was able to produce this error consistently today. 

- check out the main branch in github
- update the `config.yml` to set the validators to the `internal` mode (since there's the bug #567)
- start up inferno with `docker-compose up`.
- Once everything is up and running, go to the inferno page in chrome in two different tabs. Do all the following steps in both tabs
- select the preset tests **Inferno Reference Server (US Core v3.1.1)**
- click **Begin**
- Next to **US Core v3.1.1 Profiles** click **Run All**
- In the pop up window, click **Execute**
- wait

Within a few minutes the server should die with the message **ActiveRecord::StatementInvalid: SQLite3::BusyException: database is locked**

**Expected behavior**

We should be able to run tests in parallel. We deploy inferno as a shared service that any one of a few dozen developers can call throughout the day to test their FHIR endpoints. We need to be able to handle at least a few dozen tests executions running in parallel.

**Actual behavior**

Server dies

```
inferno_1            | 172.21.0.1 - - [24/May/2021:21:12:20 +0000] "POST /inferno/iMAyZx2p5o/test_sets/ad_hoc_testing/sequence_result HTTP/1.1" 200 - 17.7825                                                                                                                                                                                                     [25/11578]
inferno_1            | bundler: failed to load command: rackup (/usr/local/bundle/bin/rackup)
[...]

Issue #566, opened 2021-05-24
Inferno server crashing frequently from database issues


nathanloyer:
**Subject of the issue**

We maintain an inferno deployment in AWS that many teams in our company can use to validate the FHIR resources they are working on. In the latest versions of inferno, we get errors related to accessing the database every 10-15 minutes when people are actively using it. It appears to be an issue with not maintaining database locks or releasing them correctly. I'm assuming this issue was introduced when the switch to ActiveRecord was made. I've noticed that we don't have this issue in Inferno Program Edition.

**Your environment**
* Edition of inferno (Community or Program): Community
* Version of inferno: 2.12 and latest/main
* Which browser and version(s) is the bug present on?: Latest Chrome version as of 5/24/2021

Our deployment uses the docker images you are publishing to docker hub, then we copy in our configuration files, custom internal certificates needed to auth with internal endpoints, terminology validators, and some custom modules that we have written. We then publish this docker image to our internal artifactory and deploy it as an ECS service in AWS.

**Steps to reproduce**

Usually after running the US Core v3.1.1 2 to 6 times the inferno server restarts and the error logs that are attached are available. I've been reproducing by running the entire US Core test suite.

* The selected test suite: **US Core 3.1.1**
* The URL of the FHIR server being tested: **This is an internal server for our company that you don't have access to.**
* The Test ID, if it's a problem with a specific test.: **Doesn't seem to be related to a specific test.**

**Expected behavior**

We should be able to keep our inferno service deployed and have it stay online for months, instead it is crashing after a few minutes when it is being used.

**Actual behavior**

With version 2.12 I get this error:

[logs1.txt](https://github.com/onc-healthit/inferno/files/6534347/logs1.txt)

I tried working around this by updating the settings on the db config yaml file to increase the number of connections and timeout, but I got the error message below instead. I also tried updating to the latest tag from docker, which still gave me the first error message above. I also tried patching in [these changes](https://github.com/onc-healthit/inferno-program/pull/285/files) from inferno program edition because it seemed related. I still got the second error message below when using the latest version with that patch from Inferno Program Edition.

[logs2.txt](https://github.com/onc-healthit/inferno/files/6534348/logs2.txt)

Jammjammjamm:
This should be resolved by the 2.13.0 release. Please reopen this issue if it remains a problem.

Issue #511, opened 2020-12-08
Support databases other than sqlite


nathanloyer:
My team is looking at deploying inferno so we can use it to test various dev and production instances of our FHIR server. We want the test results to be persistent and for the inferno servers to be highly available, so relying on a database that is just storing data in a file in the local device storage is not going to be a viable solution. If the load balancer directs us to a different instance of inferno, trying to view the results will result in a 404. The test results also disappear when the service restarts.

In order to have a HA service, we need to be able to connect to a remote database. I've made some changes to support using postgres in my local copy of the code, and it is mostly working when just switching to the dm-postgres-adapter. However, I did have to make several updates to the data types in the data model classes because the strings that were being stored were not large enough. I can hack it together, but it would be nice to have support for this. I've probably missed various other data models that would have to be updated to support all of the test cases that my fhir server currently does not support.

As far as which databases to support, postgres is my personal preference, but any database that AWS RDS supports would be acceptable. Once one remote database is supported with the correct data types on the data models, it should be simple with the DataMapper framework to support several others.

nathanloyer:
I just noticed that your development branch has removed the datamapper framework. My request remains the same, but I'm not sure how easy the new framework is to connect to different data sources.

Jammjammjamm:
We replaced datamapper with ActiveRecord because datamapper is no longer maintained. Swapping out sqlite for postgres should require the following steps:

- Adding the `pg` gem to the `Gemfile`.
- Replacing `db/config.yml`.
- Making the schema changes you mentioned. You would create and run [migrations](https://guides.rubyonrails.org/active_record_migrations.html) to do this with ActiveRecord.

This is a skeleton `config.yml` file for postgres:
```
# PostgreSQL. Versions 9.3 and up are supported.
#
# Install the pg driver:
#   gem install pg
# On macOS with Homebrew:
#   gem install pg -- --with-pg-config=/usr/local/bin/pg_config
# On macOS with MacPorts:
#   gem install pg -- --with-pg-config=/opt/local/lib/postgresql84/bin/pg_config
# On Windows:
#   gem install pg
#       Choose the win32 build.
#       Install PostgreSQL and put its /bin directory on your path.
#
# Configure Using Gemfile
# gem 'pg'
#
default: &default
  adapter: postgresql
  encoding: unicode
  # For details on connection pooling, see Rails configuration guide
  # https://guides.rubyonrails.org/configuring.html#database-pooling
  pool: 5

development:
  <<: *default
[...]

PR #570, opened 2021-06-09
FI-1248: Migrate postgres config from program

Jammjammjamm:
# Summary
This branch migrates all of the postgres configuration from program edition. This should address #511, #566 and #568.

## New behavior
This should fix the sqlite-related db errors.

## Testing guidance
The README has instructions for running inferno with postgres. We need to make sure that everything is still working as expected.


arscan (approved):
Both configurations work fine for me, though I had to delete the data directory for the sqlite version.

PR #573, opened 2021-06-18
Switch from only POSTing StructureDefinitions to the validator service, to POSTing all conformance-related resource types

okeefm:
# Summary
This PR makes it so that `lib/app/utils/startup_tasks.rb` will upload resources from included IGs that match any of the Conformance resource types from https://www.hl7.org/fhir/conformance-module.html

Previously, it would only load `StructureDefinition` resources, which left out a number of the desired artifacts

## New behavior
* Added a `CONFORMANCE_RESOURCE_TYPES` array constant to `Inferno::StartupTasks`
* Load a resource if its type matches any in the above constant

## Testing guidance
Run Inferno with a local validator wrapper, and check that the number of `POST /profile` calls in the logs matches the number of resources for each IG that should be uploaded.

### DIFF
--- a/lib/app/utils/startup_tasks.rb
+++ b/lib/app/utils/startup_tasks.rb
@@ -8,6 +8,20 @@ require_relative '../models/module'
 module Inferno
   module StartupTasks
     class << self
+
+      CONFORMANCE_RESOURCE_TYPES = [
+        'ImplementationGuide',
+        'StructureDefinition',
+        'ValueSet',
+        'ConceptMap',
+        'CodeSystem',
+        'CapabilityStatement',
+        'OperationDefinition',
+        'SearchParameter',
+        'CompartmentDefinition',
+        'ElementDefinition'
+      ].freeze
+
       def run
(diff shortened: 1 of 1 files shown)


lib/app/utils/startup_tasks.rb
```diff
@@ -8,6 +8,20 @@
...
   module StartupTasks
     class << self
+
+      CONFORMANCE_RESOURCE_TYPES = [
+        'ImplementationGuide',
+        'StructureDefinition',
+        'ValueSet',
+        'ConceptMap',
+        'CodeSystem',
+        'CapabilityStatement',
```
Jammjammjamm, line 18: I don't know what `CapabilityStatement`, `OperationDefinition`, `SearchParameter`, or `CompartmentDefinition` would play in resource validation.

Jammjammjamm (approved):
This seems fine (once the linting error is fixed).

radamson (approved):
Tested it out and could see more artifacts being posted.

PR #574, opened 2021-06-25
Validate transactionTime in scheduling links manifest.

arscan:
# Summary

Validate that the transactionTime in the scheduling links manifest is a valid date and that it has a timezone offset as required in the spec.

See https://chat.fhir.org/#narrow/stream/281612-smart.2Fscheduling-links/topic/Publisher.3A.20Rite-Aid/near/243572426

## New behavior

Will fail if the transactionTime is not a valid time and is missing the timezone.

## Code changes

Updated one of the scheduling links tests and the tests of them.

## Testing guidance

It is difficult to mock up a failure case (hopefully the unit tests are good enough for that), but validate against valid cases to ensure it doesn't introduce incorrect failures.

### DIFF
--- a/lib/modules/smart_scheduling_links/smart_scheduling_links_basic_sequence.rb
+++ b/lib/modules/smart_scheduling_links/smart_scheduling_links_basic_sequence.rb
@@ -488,6 +488,14 @@ module Inferno
           assert_valid_http_uri file['url']
         end
 
+        begin
+          Date.iso8601(@manifest['transactionTime'])
+          has_timezone = /(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))/.match(@manifest['transactionTime']).present?
+          assert has_timezone, "Manifest transactionTime '#{has_timezone}' must contain a timezone offset or end in 'Z' for UTC."
+        rescue RuntimeError, ArgumentError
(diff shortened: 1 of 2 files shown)


lib/modules/smart_scheduling_links/smart_scheduling_links_basic_sequence.rb
```diff
@@ -488,6 +488,14 @@ def get_file(file, use_token: true)
           assert_valid_http_uri file['url']
         end
 
+        begin
+          Date.iso8601(@manifest['transactionTime'])
+          has_timezone = /(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))/.match(@manifest['transactionTime']).present?
+          assert has_timezone, "Manifest transactionTime '#{has_timezone}' must contain a timezone offset or end in 'Z' for UTC."
+        rescue RuntimeError, ArgumentError
```
Jammjammjamm, line 495: `TypeError` and `DateError` are the two types of errors I was able to get here. They both inherit from `StandardError`, so you could just rescue that.

```
> Date.iso8601('abc')
Date::Error (invalid date)
> Date.iso8601(123)
TypeError (no implicit conversion of Integer into String)
> Date::Error.ancestors
=> [Date::Error, ArgumentError, StandardError, Exception, Object, Kernel, BasicObject]
> TypeError.ancestors
=> [TypeError, StandardError, Exception, Object, Kernel, BasicObject]
```

PR #580, opened 2021-08-04
FI-1292/1293 Add missing index and allow dynamic content in db config

Jammjammjamm:
This branch adds the missing index to join table to fix performance issues as the table size increases.

Also allows erb to be used in the database configuration so that values can be loaded from environment variables.

## Testing guidance
This should work:
```
docker-compose -f docker-compose.postgres.yml build
docker-compose -f docker-compose.postgres.yml up
```
You can validate that the environment variables work by adding something like this to the inferno service in `docker-compose.postgres.yml`, which will cause inferno to fail to connect to the database.
```
    environment:
      INFERNO_DB_USERNAME: xyz
```

### DIFF
--- a/db/config.postgres.yml
+++ b/db/config.postgres.yml
@@ -13,6 +13,7 @@ production:
   adapter: postgresql
   database: <%= ENV.fetch('INFERNO_DB_NAME', 'inferno_community_prod') %>
   username: <%= ENV.fetch('INFERNO_DB_USERNAME', 'postgres') %>
+  password: <%= ENV.fetch('INFERNO_DB_PASSWORD', '') %>
   host: <%= ENV.fetch('INFERNO_DB_HOST', 'db') %>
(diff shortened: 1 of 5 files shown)


nathanloyer:
Hi @Jammjammjamm, thanks for working on this, I really appreciate it. Could you add support for setting the postgres password as well? I will need that in order to revert all the customizations I made to our inferno deployments.

Is there a PR to do this in inferno program as well?

Jammjammjamm:
I'll look into adding password, just need to make sure it doesn't break passwordless configs.

Program PR: https://github.com/onc-healthit/inferno-program/pull/334

arscan (approved):

yunwwang (approved):

nathanloyer:
@Jammjammjamm @yunwwang Hey guys, was wondering if you could kick off a release of community edition so I can pick up these changes. I was just looking at removing my workarounds/hacks to use these env vars and noticed there hasn't been a release since June.

PR #583, opened 2021-08-13
FI-1305: Fix race condition

Jammjammjamm:
Updates fhir_client to fix the race condition in reference resolution tests.


radamson (approved):
Tested it out and everything still works.

yunwwang (approved):

